PR #1, opened 2020-03-28
Using installaion event and Branch Source SCM done!

XiongKezhi:
My further attempt explained [here](https://github.com/XiongKezhi/checks-api-plugin#other-attempts), it deprecates the checks events and uses the installation events from GitHub and GitHub Branch Source SCM. I made it a PR to avoid conflicts with [the basic workflow](https://github.com/XiongKezhi/checks-api-plugin#basic-workflow). I will eventually merge this PR after the GSoC mentors from Jenkins have reviewed or even accepted my proposal.


pom.xml
```diff
@@ -45,15 +45,21 @@
...
-            <groupId>org.kohsuke</groupId>
+            <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-api</artifactId>
-            <version>1.108</version>
+            <version>1.111</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>github-branch-source</artifactId>
+            <version>2.6.0</version>
```
timja, line 56: shouldn't this be 2.7.1?

pom.xml
```diff
@@ -45,15 +45,21 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-step-api</artifactId>
+            <version>2.19</version>
```
timja, line 61: we should pull in versions from the plugin 'bom' where possible
https://github.com/jenkinsci/bom

this artifact is included

pom.xml
```diff
@@ -100,6 +106,11 @@
                 <artifactId>commons-io</artifactId>
                 <version>2.6</version>
             </dependency>
+            <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-databind</artifactId>
```
timja, line 111: the jackson2-api plugin should be used here:
https://plugins.jenkins.io/jackson2-api/

src/main/java/io/jenkins/plugins/CheckGHEventSubscriber.java
```diff
@@ -59,16 +60,41 @@ protected boolean isApplicable(@Nullable Item project) {
...
+            case INSTALLATION_REPOSITORIES:
+                LOGGER.log(Level.FINE, "Received Installation Repositories Event...");
+                updateInstallations(event); // add (repository_name, installation_id) pair
                 break;
             default:
                 throw new IllegalStateException("Unexpected value: " + event);
         }
     }
+
+    /**
```
timja, line 72: this java doc isn't useful imo, it just repeats the method name and parameter,

it's a private method and self describing so I would remove it

src/main/java/io/jenkins/plugins/CheckGHEventSubscriber.java
```diff
@@ -59,16 +60,41 @@ protected boolean isApplicable(@Nullable Item project) {
...
+     * Update installations
+     *
+     * @param event installation repositories event
+     */
+    private void updateInstallations(GHSubscriberEvent event) {
+        JsonNode payload = null;
+        try {
+            payload = new ObjectMapper().readTree(event.getPayload());
+        } catch (IOException e) {
+            e.printStackTrace();
```
timja, line 82: nit: should use a logger, can be improved later though

src/main/java/io/jenkins/plugins/CheckGHEventSubscriber.java
```diff
@@ -59,16 +60,41 @@ protected boolean isApplicable(@Nullable Item project) {
...
+        }
+
+        if (payload == null)
```
timja, line 85: is this legitimate? should it be logged?

src/main/java/io/jenkins/plugins/CheckGHEventSubscriber.java
```diff
@@ -59,16 +60,41 @@ protected boolean isApplicable(@Nullable Item project) {
...
+
+    /**
+     * Update installations
+     *
+     * @param event installation repositories event
+     */
+    private void updateInstallations(GHSubscriberEvent event) {
+        JsonNode payload = null;
+        try {
+            payload = new ObjectMapper().readTree(event.getPayload());
```
timja, line 80: should we map this to an object instead of parsing json by hand?

src/main/java/io/jenkins/plugins/CheckGHEventSubscriber.java
```diff
@@ -59,16 +60,41 @@ protected boolean isApplicable(@Nullable Item project) {
...
+        if (payload == null)
+            return;
+
+        long installationId = payload.get("installation").get("id").asLong();
+
+        // iterate the added repositories
+        Iterator<JsonNode> repoIter = payload.get("repositories_added").iterator();
+        while (repoIter.hasNext())
+            repoFullNameToInstallationId.put(repoIter.next().get("full_name").asText(), installationId);
+
```
timja, line 94: formatting might not be right but please use braces so that if someone adds a new line it's clear if its in the loop or not
```suggestion
        while (repoIter.hasNext()) {
            repoFullNameToInstallationId.put(repoIter.next().get("full_name").asText(), installationId);
            }

```

src/main/java/io/jenkins/plugins/CheckGHEventSubscriber.java
```diff
@@ -59,16 +60,41 @@ protected boolean isApplicable(@Nullable Item project) {
...
+        // iterate the removed repositories
+        repoIter = payload.get("repositories_removed").iterator();
+        while (repoIter.hasNext())
+            repoFullNameToInstallationId.remove(repoIter.next().get("full_name").asText());
```
timja, line 98:
```suggestion
        while (repoIter.hasNext()) {
            repoFullNameToInstallationId.remove(repoIter.next().get("full_name").asText());
            }
```

src/main/java/io/jenkins/plugins/Listener.java
```diff
@@ -1,15 +1,14 @@
...
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Date;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
 import java.util.TimeZone;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.CheckForNull;
```
timja, line 11: use the spotbugs annotations rather than findbugs, there's questionable licensing issues on these, see https://github.com/jenkinsci/jenkins/pull/4604 where it was removed from core

src/main/java/io/jenkins/plugins/Listener.java
```diff
@@ -62,50 +75,31 @@
...
+                    for (CheckRunSource runSource : CheckRunSource.all()) {
+                        long checkRunId = createCheckRun(runSource, repoFullName, headSha, token);
+                        run.addAction(new CheckRunAction(checkRunId, runSource));
                     }
-                    // Remove this event
-                    events.remove(event);
-                    break;
                 }
+            } catch (IOException | InterruptedException e) {
+                e.printStackTrace();
```
timja, line 101: nit use a logger

src/main/java/io/jenkins/plugins/CheckGHEventSubscriber.java
XiongKezhi: This is a kind of situation I don't know how to handle elegantly. If I remove these two lines, I'll get an `NP_NULL_ON_SOME_PATH_EXCEPTION` spot-bug warning. I could either just judge the `null` and return or return earlier in the `catch`, but I don't know which one is better?

timja: the reason is because you aren't throwing an exception on line 82

src/main/java/io/jenkins/plugins/CheckGHEventSubscriber.java
timja: easiest fix probably, it'll also fix the issue on line 85
```suggestion
            throw new UncheckedIOException(e);
```

XiongKezhi:
> 
> 
> Fixed the nits, but not the JSON parsing since we will eventually deprecate all lines of code related to the installation id map

CC @timja @uhafner

PR #2, opened 2020-05-23
Migrating parent pom to analysis-pom

XiongKezhi:
Migrating parent pom to [analysis-pom](https://github.com/jenkinsci/analysis-pom-plugin) for convenient code style checking.

@timja @uhafner

### DIFF
--- a/src/main/java/io/jenkins/plugins/util/GHAuthenticateHelper.java
+++ b/src/main/java/io/jenkins/plugins/util/GHAuthenticateHelper.java
@@ -24,7 +24,7 @@ public class GHAuthenticateHelper {
      * @param key   GitHub app private key
      *
      * @return      GitHub with jwt
-     * @throws IOException
+     * @throws IOException Exception
      */
     public static GitHub getGitHub(final String id, final String key) throws IOException {
         String jwtToken = JwtHelper.createJWT(id, key);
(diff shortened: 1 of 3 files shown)


src/main/java/io/jenkins/plugins/util/GHAuthenticateHelper.java
```diff
@@ -54,7 +54,7 @@ public static GHAppInstallationToken getInstallation(final GitHub gitHub, final
      * @param key               GitHub app private key
      *
      * @return                  GitHub app installation token
-     * @throws IOException
+     * @throws IOException Exception
```
timja, line 57: This description is not useful, same for the other occurrence

XiongKezhi: I just thought that part of the code would be replaced soon, so didn't pay much effort on it.

src/main/java/io/jenkins/plugins/util/GHAuthenticateHelper.java
```diff
@@ -35,11 +35,11 @@ public static GitHub getGitHub(final String id, final String key) throws IOExcep
...
      *
-     * @param gitHub    GitHub with jwt
-     * @param id        Installation id
+     * @param gitHub GitHub with jwt
+     * @param id Installation id
      *
-     * @return          GitHub app installation Token
-     * @throws IOException
+     * @return GitHub app installation Token
+     * @throws IOException if creat token failed
```
timja, line 42: Typo on create token

PR #3, opened 2020-05-23
Improve project setup

uhafner:
Add a `Jenkinsfile` so one can build the plugin locally in Jenkins: this helps to visualize the test results, the code and mutation coverage, and the static analysis results. Add a first test case so that PIT does not break the build because of zero tests.
  
Let the plugin depend on `plugin-utils-api` to get some assistance to make plugin code easier to test. Also add an Architecture test that forbids the usage of several libraries and classes.

### DIFF
--- a/etc/Jenkinsfile.local
+++ b/etc/Jenkinsfile.local
@@ -11,7 +11,6 @@ node {
         }
 
         recordIssues tools: [java(), javaDoc()], aggregatingResults: 'true', id: 'java', name: 'Java'
-        recordIssues tool: errorProne(), healthy: 1, unhealthy: 20
 
         junit testResults: '**/target/*-reports/TEST-*.xml'
 
--- a/etc/Jenkinsfile.local
+++ b/etc/Jenkinsfile.local
@@ -1,0 +1,43 @@
+node {
+    def mvnHome = tool 'mvn-default'
+
+    stage ('Checkout') {
+        checkout scm
+    }
+
+    stage ('Build and Static Analysis') {
+        withMaven(maven: 'mvn-default', mavenLocalRepo: '/var/data/m2repository', mavenOpts: '-Xmx768m -Xms512m') {
+            sh 'mvn -ntp -V -e clean verify -Dmaven.test.failure.ignore -Dgpg.skip'
+        }
+
+        recordIssues tools: [java(), javaDoc()], aggregatingResults: 'true', id: 'java', name: 'Java'
+        recordIssues tool: errorProne(), healthy: 1, unhealthy: 20
+
+        junit testResults: '**/target/*-reports/TEST-*.xml'
+
+        recordIssues tools: [checkStyle(pattern: 'target/checkstyle-result.xml'),
+            spotBugs(pattern: 'target/spotbugsXml.xml'),
+            pmdParser(pattern: 'target/pmd.xml'),
+            cpd(pattern: 'target/cpd.xml'),
+            taskScanner(highTags:'FIXME', normalTags:'TODO', includePattern: '**/*.java', excludePattern: 'target/**/*')],
+            qualityGates: [[threshold: 1, type: 'TOTAL', unstable: true]]
+    }
+
+    stage ('Line and Branch Coverage') {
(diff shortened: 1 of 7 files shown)


etc/Jenkinsfile.local
```diff
@@ -0,0 +1,38 @@
+node {
+    def mvnHome = tool 'mvn-default'
+
+    stage ('Checkout') {
+        checkout scm
+    }
+
+    stage ('Build and Static Analysis') {
+        withMaven(maven: 'mvn-default', mavenLocalRepo: '/var/data/m2repository', mavenOpts: '-Xmx768m -Xms512m') {
```
XiongKezhi, line 9: Shouldn't the maven local repository be something like '~/.m2/repository'?

src/main/java/io/jenkins/plugins/CheckGHEventSubscriber.java
```diff
@@ -51,7 +52,7 @@ protected boolean isApplicable(@Nullable Item project) {
      */
     @Override
     protected Set<GHEvent> events() {
-        return immutableEnumSet(GHEvent.INSTALLATION_REPOSITORIES, GHEvent.CHECK_RUN, GHEvent.CHECK_SUITE);
+        return new HashSet<>(Arrays.asList(GHEvent.INSTALLATION_REPOSITORIES, GHEvent.CHECK_RUN, GHEvent.CHECK_SUITE));
```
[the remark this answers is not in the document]
uhafner: Yes, using `immutableEnumSet` requires to use a separate library (Google Guava). Just to get a immutable set. This seems to be not adequate.

src/test/java/io/jenkins/plugins/ArchitectureTest.java
```diff
@@ -0,0 +1,33 @@
...
+    static final ArchRule NO_JENKINS_INSTANCE_CALL = PluginArchitectureRules.NO_JENKINS_INSTANCE_CALL;
+
+    @ArchTest
+    static final ArchRule NO_PUBLIC_TEST_CLASSES = PluginArchitectureRules.NO_PUBLIC_TEST_CLASSES;
+
+    @ArchTest
+    static final ArchRule NO_TEST_API_CALLED = ArchitectureRules.NO_TEST_API_CALLED;
+
+    @ArchTest
+    static final ArchRule NO_FORBIDDEN_PACKAGE_ACCESSED = PluginArchitectureRules.NO_FORBIDDEN_PACKAGE_ACCESSED;
```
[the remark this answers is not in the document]
uhafner: `org.apache.commons.lang` is not used in the pom, only `org.apache.commons.lang3`. These are 2 different libraries and we should not depend on the former one.

etc/Jenkinsfile.local
uhafner: Sorry, this is my local setup from https://github.com/uhafner/warnings-ng-plugin-devenv. It automatically builds on an agent. You can adapt that to fit your needs. I just wanted to give you an example...

uhafner:
If you would rather prefer single PRs for the different things please let me know.

uhafner:
Since you asked in our first meeting that you would like to use TDD (and use some advanced testing techniques) I thought it might be useful to provide you with the initial setup. Typically, such a setup consumes a lot of time until everything is working as expected.

etc/Jenkinsfile.local
XiongKezhi: I think it makes sense we use the default since a lot of people are using the default. And for custom users, the default can be overridden by `<localRepository>` in ~/.m2/settings.xml.

```suggestion
        withMaven(maven: 'mvn-default', mavenOpts: '-Xmx768m -Xms512m') {
```

XiongKezhi:
> 
> 
> Since you asked in our first meeting that you would like to use TDD (and use some advanced testing techniques) I thought it might be useful to provide you with the initial setup. Typically, such a setup consumes a lot of time until everything is working as expected. So I added https://www.archunit.org as well, so we can enforce some architecture rules.

I'm still not sure about TDD, just afraid it would be too challenging. We can dicsuss more about the test plan in our next meeting.

PR #5, opened 2020-05-27
Update github action

XiongKezhi:
Removed mac-os and included JDK11, code coverage

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,6 +1,4 @@
-# This workflow will build a Java project with Maven on all platforms
-
-name: Java CI with Maven on all platforms
+name: CI
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -13,16 +11,32 @@ jobs:
...
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven
+
       - name: Build with Maven
-        run: mvn -V -ntp clean verify --file pom.xml '-Dgpg.skip'
+        run: mvn -Penable-jacoco clean verify -B -V -ntp
```
timja, line 37: does a profile need to be added for this?

XiongKezhi: Yes, in order to generate the jacoco report. It's defined in the original jenkins-plugin pom

.github/workflows/maven.yml
```diff
@@ -13,16 +11,32 @@ jobs:
...
       - uses: actions/checkout@v2
-      - name: Set up JDK 1.8
+
+      - name: Set up JDK ${{ matrix.java }}
         uses: actions/setup-java@v1
         with:
-          java-version: 1.8
+          java-version: ${{ matrix.java }}
+
+      - name: Cache dependencies
```
uhafner, line 28: On what level is this cache working? Is this cache shared between the matrix elements?

XiongKezhi: Not shared between matrix elements, just cached for one single os's builds. See https://github.com/actions/cache#cache

uhafner: Why do we need a cache then? There is only one step on a node?

XiongKezhi: "A workflow can access and restore a cache created in the current branch, the base branch (including base branches of forked repositories), or the default branch (usually master)." from [restrictions-for-accessing-a-cache](https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache).

So if I understand this correctly, the cache can be shared among different branches, thus speed up the action.

timja: It’s to cache dependencies and make the build a bit faster

XiongKezhi: > 
> 
> Not shared between matrix elements, just cached for one single os's builds. See https://github.com/actions/cache#cache

Maybe I should say "cached for the same os's builds among actions in different branches"

PR #8, opened 2020-05-31
Use GitHubAppCrendential from github-branch-source

XiongKezhi:
- Migrate to use `GitHubAppCredential`
- Deprecate `CheckGHEventSubscriber`

Not sure if it's the right way to access credentials in Jenkins


src/main/java/io/jenkins/plugins/github/checks/CheckGHEventSubscriber.java
```diff
@@ -22,27 +13,30 @@
 import hudson.Extension;
 import hudson.model.Item;
 
+
+/**
+ * This subscriber manages the {@link org.kohsuke.github.GHEvent} CHECK_RUN and CHECK_SUITE
+ *
+ * @deprecated This plugin now does not use checks events to passively trigger jenkins builds
+ */
+@Deprecated
```
timja, line 22: this plugin has never been released so we shouldn't leave deprecated code in, can we remove it please?

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
```diff
@@ -75,12 +69,13 @@ public void onInitialize(Run run) {
                 String repoFullName = source.getRepoOwner() + "/" + source.getRepository();
                 String headSha = resolveHeadCommit(source.fetch(head, null));
 
-                // find installation id by repository
-                long installationId = subscriber.findInstallationIdByRepository(repoFullName);
-                if (installationId != 0) {
+                GitHubAppCredentials appCredentials = CredentialsProvider.findCredentialById(
+                        StringUtils.defaultIfEmpty(source.getCredentialsId(), ""),
+                        GitHubAppCredentials.class, run,
+                        URIRequirementBuilder.create().withUri(source.getApiUri()).build());
```
timja, line 75: shouldn't need this uri requirement when you know the ID

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
```diff
@@ -112,12 +107,13 @@ public void onStarted(Run run, TaskListener listener) {
                 // get repository full name
                 String repoFullName = source.getRepoOwner() + "/" + source.getRepository();
 
-                // find installation id by repository
-                long installationId = subscriber.findInstallationIdByRepository(repoFullName);
-                if (installationId != 0) {
+                GitHubAppCredentials appCredentials = CredentialsProvider.findCredentialById(
```
timja, line 110: this code is repeated many times in this file, can you extract out a method which takes the credential ID as a parameter?

src/main/java/io/jenkins/plugins/github/checks/CheckGHEventSubscriber.java
XiongKezhi: Do you think we should support re-run in the future? If that is the case, we can left it remains now.

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
XiongKezhi: Do we need to extract out a whole new class to include these checks creation/updating code instead of leaving them here in this listener?

src/main/java/io/jenkins/plugins/github/checks/CheckGHEventSubscriber.java
timja: we should definitely support re-run in the future, at least for the main build, individual checks i doubt we'll be able to

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
timja: not necessarily, but this class could do with some refactoring,

I would extract the finding of the credential to a method,

the determination of if it's a pull request to a method

the http code for contacting github should go elsewhere it's not related to listening to jobs

src/main/java/io/jenkins/plugins/github/checks/CheckGHEventSubscriber.java
XiongKezhi: Then I'll just remove the deprecation annotation here.

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
XiongKezhi: OK, but the whole refactor would be a big change, and also includes making use of the github-api library instead of raw http client. I'll start by a draft PR after this.

timja: sounds good

timja: you can still extract this specific method out in this PR as it's related to this code and it can be a private method in this class for now

XiongKezhi: OK. BTW, is it necessary to support the [`SCMRevisionImpl`](https://github.com/XiongKezhi/checks-api-plugin/blob/931d995917e96502e32e9a36b7a5773f967b6e51/src/main/java/io/jenkins/plugins/github/checks/JobListener.java#L271)? I tried some project configurations but I have never triggered such a revision.

XiongKezhi: Previously, the code to resolve head commit is just something I borrowed from github-branch-source. :)

timja: have you tried on a master branch?

XiongKezhi: No, but maybe the master branch is the one since we excluded it from the beginning: https://github.com/XiongKezhi/checks-api-plugin/blob/931d995917e96502e32e9a36b7a5773f967b6e51/src/main/java/io/jenkins/plugins/github/checks/JobListener.java#L72

timja: O probably shouldn't do that, checks aren't just for pull requests

XiongKezhi: Yes, it's the master branch.

XiongKezhi:
Seems I mess up the commits with many duplicates ones, haven't figure out what's the problem.....

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
```diff
@@ -276,4 +263,10 @@ private static String resolveHeadCommit(SCMRevision revision) throws IllegalArgu
             throw new IllegalArgumentException("did not recognize " + revision);
         }
     }
+
+    private static GitHubAppCredentials findGitHubAppCredentials(GitHubSCMSource source, Run<?, ?> run) {
+        return CredentialsProvider.findCredentialById(
+                StringUtils.defaultIfEmpty(source.getCredentialsId(), ""),
```
timja, line 269: if the credential id is null we should just return null, passing it down is weird in my opinion

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
```diff
@@ -276,4 +263,10 @@ private static String resolveHeadCommit(SCMRevision revision) throws IllegalArgu
             throw new IllegalArgumentException("did not recognize " + revision);
         }
     }
+
+    private static GitHubAppCredentials findGitHubAppCredentials(GitHubSCMSource source, Run<?, ?> run) {
```
timja, line 267: maybe add this
```suggestion
    @CheckForNull
    private static GitHubAppCredentials findGitHubAppCredentials(GitHubSCMSource source, Run<?, ?> run) {
```

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
XiongKezhi: Got it!

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
```diff
@@ -139,32 +129,30 @@ public void onStarted(Run run, TaskListener listener) {
...
-            } catch (IOException e) {
-                LOGGER.log(Level.WARNING,
-                        "Could not update check runs to COMPLETED. Message: " + e.getMessage(),
-                        LOGGER.isLoggable(Level.FINE) ? e : null);
+        if (source == null ) {
+            return; // not supported source and head
+        }
+
+        try {
+            GitHub gitHub = new GitHubBuilder().build();
```
timja, line 139: Github branch source just refactored this to add okhttp3, proxy support and rate limiting for github apps, we should adapt to that, should be very straight forward:

https://github.com/jenkinsci/github-branch-source-plugin/commit/342952d4f51a6c75a439c62b8a35a0bf08990e71

XiongKezhi: Maybe not necessary for now, since these codes will be replaced by the newly designed API. 

We can adapt these changes by then, what do you think?

timja: when we change it we should adapt these changes otherwise it won't work for some users

XiongKezhi: Made an issue https://github.com/XiongKezhi/checks-api-plugin/issues/17 for this, will remember that.

PR #13, opened 2020-06-02
API Design

XiongKezhi:
This is a follow-up for [PR 6](https://github.com/XiongKezhi/checks-api-plugin/pull/6), this PR focus more on the checks-plugin side.

This is a just draft PR for discussion.


uhafner:
It would be also helpful if you would create some tests sketches that indicate how the API will be used.

XiongKezhi:
Sorry, I checkout this branch on a feature branch and so introduced unrelated changes, I’ll close this and open a new one that only contains API changes.

PR #14, opened 2020-06-03
Draft API design

XiongKezhi:
A follow-up for [PR #6](https://github.com/XiongKezhi/checks-api-plugin/pull/6).

This PR is about how the checks plugin retrieves checks parameters from consumers and interact with GitHub.


src/main/java/io/jenkins/plugins/github/checks/ChecksPublisher.java
```diff
@@ -0,0 +1,29 @@
...
+import org.kohsuke.github.GHCheckRun;
+import org.kohsuke.github.GHCheckRun.Status;
+import org.kohsuke.github.GitHubBuilder;
+
+import io.jenkins.plugins.github.checks.api.ChecksBuilder;
+
+/*
+ * This publisher manages interactions with GitHub, including creating and updating check runs.
+ */
+class ChecksPublisher {
```
timja, line 16: Is this intended to be part of the API or internal only?

XiongKezhi: Internally, so now I set it to be packet visible. We can discuss more about the whole design in today's meeting

src/main/java/io/jenkins/plugins/github/checks/api/ChecksListener.java
```diff
@@ -0,0 +1,17 @@
...
+import org.kohsuke.accmod.restrictions.Beta;
+import hudson.ExtensionPoint;
+import hudson.model.Run;
+
+/**
+ * This listener provides extension points to notify consumers provide parameters for checks.
+ */
+@Restricted(Beta.class)
+public interface ChecksListener extends ExtensionPoint {
+    abstract void onQueued(final Run<?, ?> run, final ChecksBuilder checks, final AnnotationsBuilder annotations);
```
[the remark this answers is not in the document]
XiongKezhi: Sorry, I implemented it as an abstract class before; later I found only methods are needed, so I changed it into interface and forget to remove `abstract` keyword.

src/main/java/io/jenkins/plugins/github/checks/ChecksPublisher.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.jenkinsci.plugins.github_branch_source.GitHubAppCredentials;
+
+import io.jenkins.plugins.github.checks.api.ChecksBuilder;
+
+/*
+ * This publisher manages interactions with GitHub, including creating and updating check runs.
+ */
+public class ChecksPublisher {
+    public static void createCheckRun(ChecksBuilder checks, String token) {}
```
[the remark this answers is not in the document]
XiongKezhi: I guess it's the name "publisher" that makes it somewhat confusing, especially in Jenkins, but I made this class intend to include the interactions with GitHub.

src/main/java/io/jenkins/plugins/github/checks/ChecksPublisher.java
timja: sure,

normally you use `@Restricted(NoExternalUse.class)` in Jenkins to be very clear that no one should use the class as well

XiongKezhi: OK, will add it later

XiongKezhi:
I'll add tests later...

src/main/java/io/jenkins/plugins/github/checks/DefaultPublisher.java
```diff
@@ -0,0 +1,18 @@
...
+import io.jenkins.plugins.github.checks.api.ChecksPublisher;
+
+/**
+ * A default publisher that only provides name of a check run.
+ */
+@Extension
+@Restricted(NoExternalUse.class)
+public class DefaultPublisher implements ChecksPublisher {
+    public String getName() {
+        return "Jenkins Build";
```
timja, line 16:
```suggestion
        return "Jenkins";
```

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksClient.java
```diff
@@ -0,0 +1,103 @@
...
+import org.kohsuke.github.GHCheckRun.Conclusion;
+import org.kohsuke.github.GHCheckRun.Status;
+import org.kohsuke.github.GitHubBuilder;
+
+import io.jenkins.plugins.github.checks.api.ChecksDetails;
+
+/*
+ * A Client used to connect GitHub and manipulate check runs.
+ */
+public class GitHubChecksClient {
```
timja, line 19: restrict this so people don't use it directly for now?

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksClient.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Create a check with <code>details</code> of a specific repository and commit described in <code>details</code>.
+     *
+     * @param context
+     *         context of the check
+     * @param details
+     *         details of the check
+     */
+    public static void createCheckRun(ChecksContext context, ChecksDetails details) {
```
timja, line 30: we can probably create an instance of this class and pass the context in the constructor? rather than having all these methods static?

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
```diff
@@ -57,34 +37,10 @@
...
-                    run.addAction(new CheckRunResultAction(checkRunId, runSource));
-                }
-            }
-        } catch (IOException | InterruptedException e) {
-            LOGGER.log(Level.WARNING,
-                    "Could not create check runs. Message: " + e.getMessage(),
-                    LOGGER.isLoggable(Level.FINE) ? e : null);
+        final ChecksContext context = new ChecksContext(run);
+        for (ChecksPublisher publisher : jenkins.getExtensionsFor(ChecksPublisher.class)) {
+            ChecksDetails checks = new ChecksDetailBuilder(publisher.getName()).build();
```
timja, line 42: I prefer the syntax, `CheckDetails.builder(name).build()`

any reason you removed that?

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * By default, this method only use the <code>details</code> without any changes to create a check.
+     *
+     * @param run
+     *         the run which creates this check
+     * @param details
+     *         the checks
+     */
+    public default void publishToQueued(final Run<? ,?> run, final ChecksDetails details) {
+        GitHubChecksClient.createCheckRun(new ChecksContext(run), details);
```
timja, line 34: I don't think consumers should implement these methods at all, they shouldn't have any awareness that this is going to github, with this design they would have to add this method at the bottom of the logic.

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksClient.java
XiongKezhi: Any differences between instance this and just left it static?

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
XiongKezhi: I went through your tweaks PR again, so in `JobListener`, we no longer provide the status for individual checks like warnings-checks or code-cov-checks but provide a whole `Jenkins` check and we are not iterating the consumers but iterating implementations of publishers for like GitHub, GitLab, etc.?

If users want to publish to GitHub immediately, they just code like:
```
ChecksPublisher publisher = GitHubPublisher();
publisher.publish(context, details);
```

XiongKezhi: If my understanding right now is correct, I may misunderstand you before this today's design and will change it.

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksClient.java
timja: if it's an instance then you don't have to pass the context around on every method call, I would expect the context to not change between most calls so easier to read the code I think

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
timja: it's a static method on the CheckDetails class that constructs the builder, although you can also do it in the builder

there's many different ways to do it, 

and I've just looked at 5 different articles and seen 3 different code conventions, and can't even see the one I'm used to 🤷‍♂️

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
timja: I think it would be something like:

```
ChecksPublisher publisher = ExtensionList.lookupSingleton(ChecksPublisher.class);
publisher.publish(run, details)
```

but yes that's what I meant

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
XiongKezhi: hahaha, maybe you can change to this style as well ;)

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
XiongKezhi: But in your tweaks, what's the use of the main "Jenkins" check-in `JobListener`, also, are we just ignoring the status of consumers checks?

timja: The main Jenkins check is for the whole build, similar to the GitHub-branch-source commit status

Consumers in my tweaks were publishing immediately I just didn’t remove the code in the job listener

I was wondering if it would make sense to have both options, an immediate one and one during the build listener

XiongKezhi: I changed my code today. I didn't make the `publisher` an abstraction for different platforms, but I made the `ChecksClient` an abstraction for them. 

I still use the `publisher` as the API and add the ability for them to choose what status update they want checks-plugin to do automatically for them. But if they want something immediately like publishing the completed result, they could just call this default `publishToCompleted` directly.

I'll push the code here, just wait a minute.

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
```diff
@@ -0,0 +1,74 @@
...
+     * Complete a check.
+     *
+     * By default, this method only use the <code>details</code> without any changes to complete a check.
+     *
+     * @param run
+     *         the run which creates this check
+     * @param details
+     *         the checks
+     */
+    public default void publishToCompleted (final Run<?, ?> run, final ChecksDetails details) {
```
XiongKezhi, line 71: The consumers should just call this method if they want an immediate publish, but they should make sure `autoStatus` method won't return `ChecksStatus.Completed`, or the `JobListener` may overwrite the checks with an empty one.

src/main/java/io/jenkins/plugins/github/checks/ChecksContext.java
```diff
@@ -0,0 +1,109 @@
...
+import org.jenkinsci.plugins.github_branch_source.GitHubSCMSource;
+import org.jenkinsci.plugins.github_branch_source.PullRequestSCMRevision;
+import hudson.model.Run;
+import hudson.util.Secret;
+import jenkins.plugins.git.AbstractGitSCMSource.SCMRevisionImpl;
+import jenkins.scm.api.SCMHead;
+import jenkins.scm.api.SCMRevision;
+import jenkins.scm.api.SCMSource;
+
+public class ChecksContext {
```
XiongKezhi, line 22: Now this `ChecksContext` maybe not abstract enough, if this design works for us, some revision will be done here.

src/main/java/io/jenkins/plugins/github/checks/ChecksClient.java
```diff
@@ -0,0 +1,81 @@
...
+     * Complete a check run.
+     *
+     * @param context
+     *         the context of a check
+     * @param details
+     *         the detailed parameters of a check
+     */
+    public abstract void completeCheckRun(ChecksContext context, ChecksDetails details);
+
+    public static void createCheckRun(Run<?, ?> run, ChecksDetails details) {
```
XiongKezhi, line 55: These three methods will iterate all implementations of `ChecksClient` and use the right one to access checks API.

For example, the `GitHubChecksClient` should return true by `isApplicable()` when passing a `ChecksContext` which contains the `GitHubSCMSource` to it; and a `GitLabChecksClient` should return true by `isApplicable()` when passing a `ChecksContext` which contains the `GitLabSCMSource`.

src/main/java/io/jenkins/plugins/github/checks/ChecksContext.java
```diff
@@ -0,0 +1,109 @@
...
+    public SCMSource getSource() {
+        return source;
+    }
+
+    public Run<?, ?> getRun() {
+        return run;
+    }
+
+    @CheckForNull
+    private static GitHubSCMSource resolveSource(Run<?, ?> run) {
```
XiongKezhi, line 78: In order to make such situations testable, should I add another method that takes both  `run` and `SCMSource.SourceByItem` as parameters? Then I can mock the `SourceByItem.findSource` behavior? @uhafner @timja

XiongKezhi: Something like:
```
    static GitHubSCMSource resolveSource(Run<?, ?> run) {
        return (GitHubSCMSource) (resolveSource(run, GitHubSCMSource.class));
    }

    static <T extends SCMSource> SCMSource resolveSource(Run<?, ?> run, Class<T> type) {
        return SCMSource.SourceByItem.findSource(run.getParent());
    }
```

uhafner: You should put every of such calls behind a facade. See `JenkinsFacade` as an example. Then you can replace the facade in the tests with a stub.

src/main/java/io/jenkins/plugins/github/checks/ChecksContext.java
```diff
@@ -0,0 +1,109 @@
...
+        return (GitHubSCMSource) SCMSource.SourceByItem.findSource(run.getParent());
```
uhafner, line 79: I'm not sure if we should discuss this with a wider audience. While I added SCM blames to the warnings plugin I found it quite complicated to get the SCM for a `Run`: https://github.com/jenkinsci/forensics-api-plugin/blob/master/src/main/java/io/jenkins/plugins/forensics/util/ScmResolver.java

Is the same true for `GitHubSCMSource` as well? What if a pipeline contains 2 repositories. One in GitHub, one in GitLab. Should we support such a case? I am fine with saying, we just allow one of such providers per run, but maybe we should discuss that...

src/main/java/io/jenkins/plugins/github/checks/ChecksStatus.java
```diff
@@ -0,0 +1,7 @@
+package io.jenkins.plugins.github.checks;
+
+public enum ChecksStatus {
+    Queued,
```
uhafner, line 4: enum values are typically in `UPPER_CASE`

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksClient.java
```diff
@@ -0,0 +1,113 @@
...
+        if (StringUtils.isNoneBlank(repository, headSha, token)) {
+            try {
+                new GitHubBuilder().withAppInstallationToken(token).build()
+                        .getRepository(repository).createCheckRun(details.getName(), headSha)
+                        .withStartedAt(new Date())
+                        .withDetailsURL(details.getDetailsURL())
+                        .withStatus(Status.QUEUED)
+                        .create();
+            } catch (IOException e) {
+                LOGGER.log(Level.WARNING, "Could not create check run. Message: ", e);
```
uhafner, line 54: Wouldn't it make sense to report to the `Run`? Then the errors are visible in the console log.

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksClient.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Complete a check with <code>details</code> of a specific repository and commit described in <code>details</code>.
+     *
+     * @param context
+     *         context of the check
+     * @param details
+     *         details of the check
+     */
+    public void updateCheckRun(ChecksContext context, ChecksDetails details) {
+        String repository = context.getRepository();
```
uhafner, line 68: Extract the common code into a method

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public static class ChecksDetailBuilder {
+        private final String name;
+        private String detailsURL;
+        private String conclusion;
+        private List<Output> outputs;
+        private List<Action> actions;
+
+        public ChecksDetailBuilder(String name) {
+            this.name = name;
```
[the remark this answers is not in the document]
timja: humean readable afaik, 

needs documentation

src/main/java/io/jenkins/plugins/github/checks/ChecksContext.java
timja: we can probably start with just one, this will always get the one that the run was started with afaik.

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
XiongKezhi: A single method is better but I haven't found a way to allow a single method to report on different status:
* For warnings-plugin, one method is enough because we just report the warnings in the end and set the status to be `queued`
* But things may be different for other situations. For example, a long deployment which takes 30 or 40 minutes, maybe one consumer during the deployment wants to report different pieces of information alone the deploying, at this time, they will invoke `publishToInProgress` because this will set the status to be `in progress`. Eventually, when the deployment finishes, they will invoke `publishToCompleted` to publish the final report and set the status to be `completed`.

src/main/java/io/jenkins/plugins/github/checks/ChecksClient.java
```diff
@@ -0,0 +1,81 @@
...
+        ChecksContext context = new ChecksContext(run);
+        for (ChecksClient client: jenkins.getExtensionsFor(ChecksClient.class)) {
+            if (client.isApplicable(context)) {
+                client.createCheckRun(context, details);
+            }
+        }
+    }
+
+    public static void updateCheckRun(Run<?, ?> run, ChecksDetails details) {
```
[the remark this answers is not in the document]
XiongKezhi: See comments on `ChecksPublish` :)

src/main/java/io/jenkins/plugins/github/checks/ChecksClient.java
XiongKezhi: Yes, felt that evil today..

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
uhafner: But shouldn't we then simply use the desired status as part of the ChecksDetails object?

XiongKezhi: Initially, I just thought the `ChecksDetails` only include parameters in configurable and provided groups. But if now storing status in `CheckDetails` makes the API more friendly to consumers, it worth doing.

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
```diff
@@ -0,0 +1,74 @@
...
+import hudson.model.Run;
+
+import io.jenkins.plugins.github.checks.ChecksClient;
+import io.jenkins.plugins.github.checks.ChecksStatus;
+
+/**
+ * A publisher API for consumers to publish checks.
+ */
+@Restricted(Beta.class)
+public interface ChecksPublisher extends ExtensionPoint {
```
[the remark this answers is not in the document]
uhafner: Hmm, in the warnings plugin I do not need the latter one. I think that I always will publish directly. 
Psuedocode:
```
perform() {
    publisher = jenkins.getExtensionsFor(ChecksPublisher.class);
    builder = new DetailsBuilder();
    
    builder.setStatus(PROGRESS);
    publisher.publishChecks(run, details);

    // scan for issues and report issues

    builder.setStatus(FINISHED);
    publisher.publishChecks(run, details);
}
```

Do we already have a use case for the complex case you mentioned?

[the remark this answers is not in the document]
XiongKezhi: I just thought it would be a good feature to automatically set status like `pending` for consumers since it will be harder for consumers to do that themselves when the job is still in the queue. And for the use case, I believe users want to see how many checks are made by Jenkins immediately after they push their code, not nothing at the beginning, and wait a while and see several checks in progress already.

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        public ChecksDetailBuilder withDetailsURL(final String detailsURL) {
```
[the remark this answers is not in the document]
XiongKezhi: Maybe just
```
        @NonNull
        public ChecksDetailBuilder withDetailsURL(@Nullable final String detailsURL) 
            this.detailsURL = detailsURL;
            return this;
        }
```
Since from the standpoint of the user, `detailsURL` can be set null which means they don't want to provide detail URL. But we need to fix the null with empty string later when `build()`

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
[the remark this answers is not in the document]
timja: Not all checks are necessarily going to happen though, not sure if it makes sense for any of the others checks to be set until required

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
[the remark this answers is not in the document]
timja: Why do we need to set an empty string?

And they shouldn’t really call this builder with null not that it really causes any harm

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -0,0 +1,77 @@
...
+            return this;
+        }
+
+        public ChecksDetailBuilder withActions(final List<Action> actions) {
+            this.actions = actions;
+            return this;
+        }
+
+        public ChecksDetails build() {
+            return new ChecksDetails(name, detailsURL, conclusion, outputs, actions);
```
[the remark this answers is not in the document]
timja: Isn’t the only required value name? And that’s set in the constructor of the builder?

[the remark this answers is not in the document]
XiongKezhi: If we want to decrease the three publish methods to a single one, then the status will be added here. And when the status been added here, other situations may occur, like: status must be `completed` when the conclusion is provided.

I'm refactoring this part today, and I wonder if we should return a `Collections.emptyList()` to avoid null when the users don't set `outputs` and `actions`

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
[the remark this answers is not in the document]
XiongKezhi: > Not all checks are necessarily going to happen though

That's a problem, so like a pipeline script which has never invoke checks reporting, then the checks we automatically set for them will last forever...

Should we just leave the `JobListener` alone for now? I have no idea now, it seems each design can't make full use of all checks' functionalities :(

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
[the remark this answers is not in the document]
XiongKezhi: set an empty string just try to avoid return the `null` when `getDetailsURL()`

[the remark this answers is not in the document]
uhafner: If a consumer does not want to show a details URL then the consumer should simply not invoke this method. If a consumer decides to invoke the method, then the parameter should be valid, i.e. a non empty string. You can catch that later before you build the object whether this method has been invoked or not.

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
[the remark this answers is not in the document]
uhafner: My point is: if someone invokes this method, then the parameters must be valid. If a parameter is optional then the consumer should not be forced to call with `null` or an empty list, the consumer should simply not invoke the builder method.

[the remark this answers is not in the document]
timja: yes it makes set to default lists to empty and not null, you can just do that by defaulting the field to empty list, and using `requireNonNull` in the builder to ensure no one passes null to it

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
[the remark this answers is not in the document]
timja: i would just leave the job listener for the main build for now

src/main/java/io/jenkins/plugins/github/checks/ChecksContext.java
XiongKezhi: I use the `ChecksContext` to store the `SCMSource`, `head` temporarily for a check since the resolve process really takes some time in my debugging experience. But one question is, should I simply make those static resolve methods non-static or extract them into a new class may be `ContextResolver`, which one is better?

uhafner: In order to stub these calls they need to be part of a replaceable object. You can either use a static nested class or a standalone class, it does not really matter.

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
XiongKezhi: now I saw Guava uses this style :)
https://github.com/google/guava/blob/a1e9a0bd1260a657f7502e5b81d38dfb818cd1d4/guava/src/com/google/common/collect/ImmutableMap.java#L184

pom.xml
```diff
@@ -62,55 +62,118 @@
...
-        <version>${commons-lang3.version}</version>
+        <groupId>org.apache.httpcomponents</groupId>
+        <artifactId>httpclient</artifactId>
+        <version>4.5.12</version>
+      </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-core</artifactId>
+        <version>2.10.2</version>
+        <scope>test</scope>
```
timja, line 95: I don’t think we should be defining scope in dependency management https://stackoverflow.com/questions/15221299/dependencymanagement-and-scope

pom.xml
```diff
@@ -62,55 +62,118 @@
...
+      <dependency>
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-server</artifactId>
+        <version>9.4.26.v20200117</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-http</artifactId>
+        <version>9.4.26.v20200117</version>
```
timja, line 118: Extract this out to a property as it’s used for multiple jetty dependencies

pom.xml
```diff
@@ -62,55 +62,118 @@
...
+      <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-core</artifactId>
+        <version>2.10.2</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-databind</artifactId>
+        <version>2.10.2</version>
```
timja, line 100: Extract version

pom.xml
XiongKezhi: Now I know why moving these to the dependency management section solves the enforce problems: the enforced provided scope in this section. So if it would be better if I move them back to `<dependencies>` section and use the `provided` scope?

XiongKezhi: Oops, maybe just `test` scope is fine. I may just used the wrong version that causes those enforce problems...

PR #15, opened 2020-06-03
API design tweaks for immediate publish

timja:


src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
```diff
@@ -0,0 +1,19 @@
...
+import io.jenkins.plugins.github.checks.api.CheckEvent;
+import io.jenkins.plugins.github.checks.api.ChecksPublisher;
+
+/**
+ * Publishes result to GitHub immediately
+ */
+public class GitHubChecksPublisher implements ChecksPublisher {
+    
+    @Override
+    public void publish(Run<?, ?> run, CheckEvent checks, AnnotationsBuilder annotations) {
```
XiongKezhi, line 14: This method is easy to use for our plugin since we can access the run `JobListener`, but how can the users provide the run? The users invoke this `publish` method in (say) [`perform`](https://javadoc.jenkins-ci.org/jenkins/tasks/SimpleBuildStep.html#perform-hudson.model.Run-hudson.FilePath-hudson.Launcher-hudson.model.TaskListener-) method in the `SimpleBuildStep`?

timja: yes exactly, they have access to the run from the perform parameters

XiongKezhi: great, a big fix in design👍🏻

uhafner: I do not understand this workflow. Is `publish` called by the Checks API Plugin? Where do I get all those parameters in line 15? These should be provided by the Checks API. 

And where do I build my messages here? In the previous API call we had the builder as parameter where the client could build the message step by step.

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
```diff
@@ -79,12 +85,12 @@ public void onInitialize(Run run) {
                 // create token
                 String token = Secret.toString(appCredentials.getPassword());
 
-                for (ChecksListener listener : new JenkinsFacade().getExtensionsFor(ChecksListener.class)) {
-                    ChecksBuilder checks = new ChecksBuilder(listener.getName());
+                for (ChecksPublisher listener : new JenkinsFacade().getExtensionsFor(ChecksPublisher.class)) {
```
uhafner, line 88: Unrelated to the PR: the Jenkins Facade should be a field that can be replaced in the constructor by a stub. 

Example: https://github.com/jenkinsci/warnings-ng-plugin/blob/master/plugin/src/main/java/io/jenkins/plugins/analysis/core/util/ModelValidation.java#L43

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
```diff
@@ -117,15 +123,14 @@ public void onStarted(Run run, TaskListener listener) {
                 // create token
                 String token = Secret.toString(appCredentials.getPassword());
 
-                for (ChecksListener checksListener : new JenkinsFacade().getExtensionsFor(ChecksListener.class)) {
-                    ChecksBuilder checks = new ChecksBuilder(checksListener.getName());
+                for (ChecksPublisher checksPublisher : new JenkinsFacade().getExtensionsFor(ChecksPublisher.class)) {
+                    CheckEvent checks = CheckEvent.builder(BUILD_CHECK_NAME).build();
```
uhafner, line 127: Shouldn't the name come from the consumer?

src/main/java/io/jenkins/plugins/github/checks/api/CheckEvent.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public List<Action> getActions() {
+        return actions;
+    }
+
+    public static Builder builder(String name) {
+        return new Builder(name);
+    }
+    
+    public static class Builder {
```
uhafner, line 44: Shouldn't this builder be given to the consumer, so that the consumer can define all these values?

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
timja: The way I’ve done it here this JobListener is just for the main build.

Not for checks to be contributed

src/main/java/io/jenkins/plugins/github/checks/api/CheckEvent.java
timja: The consumer can construct is with CheckEvent.builder(“SpotBugs”);

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
XiongKezhi: I'm trying to use a `CheckRunContext` whose constructer take `Run` as a parameter and will include some resolve method for these parameters, I'll post that code later.

uhafner: I see. So you simply inverted the control: Not the checks plugin is calling the consumers but the consumers are calling the plugin. This makes the logic in the Checks API much simpler. And consumers can decide on their own, when to publish results. 

(Then we can not react on events (queue, job started), but this is something I do not need in the warnings plugin anyway.)

XiongKezhi: That's something I'm doing, I'm trying to decouple those checks build steps from `JobListener` and maybe unify the interface for both checks plugins' control and consumers' control.

timja: Possibly we could also allow users to hook into our job listener as well? But not sure if needed.

uhafner: Since this code is called by the consumer we can also make the annotations part of the checks "event" object. 

I'm not sure if `Event` is a good name, actually it is the data model for a pull request comment. (Or in GitLab a merge request).

timja: it's not just pull requests, you can also have checks on master as well.

but yes naming is TBC

XiongKezhi: Any ideas on the name? How about `ChecksParameters` or `ChecksDetials`?

timja: `ChecksRunDetails` or `CheckDetails` would get by vote I think

uhafner: `CheckDetails` sounds very good (though it depends on the GitHub naming). 

I think it is hard to find a generic name that indicates that we can comment pull or merge requests (or even just branch builds) in GitHub, GitLab, or whatever.

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
```diff
@@ -117,15 +123,14 @@ public void onStarted(Run run, TaskListener listener) {
                 // create token
                 String token = Secret.toString(appCredentials.getPassword());
 
-                for (ChecksListener checksListener : new JenkinsFacade().getExtensionsFor(ChecksListener.class)) {
-                    ChecksBuilder checks = new ChecksBuilder(checksListener.getName());
+                for (ChecksPublisher checksPublisher : new JenkinsFacade().getExtensionsFor(ChecksPublisher.class)) {
```
XiongKezhi, line 126: Maybe there's no need to iterate all publishers at all since we can decide to use which publisher based on the type of `SCMSource`, e.g. if the run uses `GitHubSCMSource`, we just use `GitHubPublisher` directly.

PR #19, opened 2020-06-07
Improve checks detail

XiongKezhi:
Currently use `Optional` to handle `null`, but this can be changed if necessary.


src/main/java/io/jenkins/plugins/github/checks/GitHubChecksClient.java
```diff
@@ -47,7 +47,8 @@ public void createCheckRun(ChecksContext context, ChecksDetails details) {
                 new GitHubBuilder().withAppInstallationToken(token).build()
                         .getRepository(repository).createCheckRun(details.getName(), headSha)
                         .withStartedAt(new Date())
-                        .withDetailsURL(details.getDetailsURL())
+                        .withDetailsURL(details.getDetailsURL().isPresent() ?
```
timja, line 50: use `orElse`
```suggestion
                        .withDetailsURL(details.getDetailsURL().orElse(context.getRun().getUrl())
```

uhafner: This depends a little bit on the answer of https://github.com/XiongKezhi/checks-api-plugin/pull/14#discussion_r436266916.

If we really need the `Context` type, then this is the way to go. I would suggest one minor thing ([tell don't ask](https://martinfowler.com/bliki/TellDontAsk.html)): 

```suggestion
                        .withDetailsURL(details.getDetailsURL().orElse(context.getDetailsUrl()))
```

If we can embed the `Context` information into the `Details` it would simply resolve to

```suggestion
                        .withDetailsURL(details.getDetailsURL())
```

and the `Details` instance handles everything for us

src/test/java/io/jenkins/plugins/github/checks/api/ChecksDetailsTest.java
```diff
@@ -0,0 +1,90 @@
...
+import io.jenkins.plugins.github.checks.ChecksConclusion;
+import io.jenkins.plugins.github.checks.ChecksStatus;
+import io.jenkins.plugins.github.checks.api.ChecksDetails.ChecksDetailsBuilder;
+
+import static org.assertj.core.api.Assertions.*;
+
+/**
+ * Tests the class {@link ChecksDetails}.
+ */
+class ChecksDetailsTest {
```
uhafner, line 17: 🎉

src/test/java/io/jenkins/plugins/github/checks/api/ChecksDetailsTest.java
```diff
@@ -0,0 +1,90 @@
...
+    private final String CHECK_NAME = "Jenkins";
+
+    @Test
+    void shouldCreateWhenBuildOnlyWithConstructor() {
+        ChecksDetails details = new ChecksDetailsBuilder(CHECK_NAME, ChecksStatus.QUEUED)
+                .build();
+
+        assertThat(details.getName()).isEqualTo(CHECK_NAME);
```
uhafner, line 25: See https://github.com/XiongKezhi/checks-api-plugin/pull/20/commits/93a9b04149948f618e5ef8b218e0c5ade16f49b0

src/test/java/io/jenkins/plugins/github/checks/api/ChecksDetailsTest.java
```diff
@@ -0,0 +1,90 @@
...
+                .withOutputs(outputs)
+                .withActions(actions);
+
+        ChecksDetails details = builder.build();
+        assertThat(details.getName()).isEqualTo(CHECK_NAME);
+        assertThat(details.getStatus()).isEqualTo(ChecksStatus.COMPLETED);
+        assertThat(details.getDetailsURL().isPresent()).isTrue();
+        assertThat(details.getDetailsURL().get()).isEqualTo(detailsURL);
+        assertThat(details.getOutputs().isPresent()).isTrue();
+        assertThat(details.getOutputs().get().size()).isEqualTo(2);
```
uhafner, line 55: AssertJ heavily supports the tell don't ask principle 😄 

```suggestion
        assertThat(details.getOutputs()).hasSize(2);
```

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -1,76 +1,160 @@
...
-        public ChecksDetailBuilder(String name) {
+        public ChecksDetailsBuilder(@NonNull final String name, @NonNull final ChecksStatus status)
+                throws IllegalArgumentException{
+            Objects.requireNonNull(status);
+            if (StringUtils.isBlank(name)) {
+                throw new IllegalArgumentException("check name should not be blank");
+            }
+
             this.name = name;
+            this.status = status;
```
uhafner, line 111:
```suggestion
            this.status = Objects.requireNonNull(status);
```

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -1,76 +1,160 @@
...
         private final String name;
+        private final ChecksStatus status;
         private String detailsURL;
-        private String conclusion;
+        private ChecksConclusion conclusion;
         private List<Output> outputs;
         private List<Action> actions;
 
-        public ChecksDetailBuilder(String name) {
+        public ChecksDetailsBuilder(@NonNull final String name, @NonNull final ChecksStatus status)
```
uhafner, line 103: If we decide to use `NunNull` as default, we should rather declare it on the package.:
https://github.com/jenkinsci/analysis-model/blob/master/src/main/java/edu/hm/hafner/analysis/package-info.java

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -1,76 +1,160 @@
...
+        return Optional.ofNullable(actions);
+    }
+
+    public static class ChecksDetailsBuilder {
         private final String name;
+        private final ChecksStatus status;
         private String detailsURL;
-        private String conclusion;
+        private ChecksConclusion conclusion;
         private List<Output> outputs;
```
uhafner, line 100:
```suggestion
        private List<Output> outputs = new ArrayList<>();
```

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksClient.java
XiongKezhi: Embed `Context` in `Details` is a good idea, but one problem is that we then may need to add `run` as another parameter for the constructor, is that ok?

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
XiongKezhi: Maybe better use `Collections.emptyList()`?

uhafner: Yes! Since you replace the collection anyway in `withOutput` this will be simpler.

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksClient.java
uhafner: For the immediate publishing we need to do something like (see https://github.com/XiongKezhi/checks-api-plugin/pull/14/files#r436286430)

```
Run run = ...;
// find a Publisher for our SCM provider
ChecksPublisher publisher = ChecksPublisherFactory.find(run); 

// or if we need the SCM as well
// ChecksPublisher publisher = ChecksPublisherFactory.find(run, scm);

DetailsBuilder builder = new DetailsBuilder("CheckStyle", run);

// Or the publisher sets the run in a factory method
// or DetailsBuilder builder = publisher.createDetailsBuilder("Check style")

// Or the publisher sets the run after the publish method
    
builder.withStatus(PROGRESS);

publisher.publishChecks(run, builder.build());
```

So in this case the run must be available anyway. We can retrieve from the publisher or we can add it as parameter. I'm not sure how we handle the second case (queued and therefore no run available yet).

XiongKezhi: But what about the `name` duplication problem? Before I was thinking `ChecksPublisher` to be extended by consumers and they will implement the `getName()`, so we know all the consumer checks names, but now, the control is in the hand of the users, how can we solve that?

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -1,76 +1,147 @@
...
+     * Returns the status of a check.
+     *
+     * @return {@link ChecksStatus}, one of {@code QUEUED}, {@code IN_PROGRESS}, {@code COMPLETED}
+     */
+    public ChecksStatus getStatus() {
+        return status;
+    }
+
+    /**
+     * Returns the url of a site with full details of the check.
```
timja, line 49: More detail would be good with examples, this will default to the Jenkins build url if not supplied won’t it?

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -1,76 +1,147 @@
...
         this.name = name;
+        this.status = status;
         this.detailsURL = detailsURL;
         this.conclusion = conclusion;
         this.outputs = outputs;
         this.actions = actions;
     }
 
+    /**
+     * Returns the name of a check.
```
timja, line 31: Needs to be clear this goes in GitHub UI and that users will see it, preferably giving an example as well

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -1,76 +1,147 @@
...
+     *
+     * @return the string representing the url of a site
+     */
     public String getDetailsURL() {
         return detailsURL;
     }
 
-    public String getConclusion() {
+    /**
+     * Returns the conclusion of a check.
```
timja, line 58: I don’t think we need to prefix all of these docs with returns the

XiongKezhi: Just following the coding style: https://github.com/uhafner/codingstyle/blob/master/doc/Kommentare.md#javadoc. It's in German, so I translate it in Google, I hope I didn't understand it wrong @uhafner 

Also, the Java standard library used this style of Java Doc :)

XiongKezhi:
> 
> 
> Thanks for adding the docs, a few suggestions

I'll add details to the docs of `with` methods in the builder, I guess those are more widely used by users.

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
uhafner: Yes, I took this convention from the book https://www.cambridge.org/core/books/elements-of-java-style/C07D748D3C8108468EDD7807AF693660: it helps non-native speakers to write good JavaDoc comments (wording, grammar, etc).

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksClient.java
uhafner: I don't understand why we do not need a factory method to obtain the correct publisher? Is there another way to get the publisher? The publisher is required to invoke the `publishChecks` method.  How do you obtain it currently?

timja:
> > Thanks for adding the docs, a few suggestions
> 
> I'll add details to the docs of `with` methods in the builder, I guess those are more widely used by users.

Yes definitely, users will likely only ever use the builders

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -1,76 +1,211 @@
...
-    public static class ChecksDetailBuilder {
+    public static class ChecksDetailsBuilder {
         private final String name;
+        private final ChecksStatus status;
         private String detailsURL;
-        private String conclusion;
-        private List<Output> outputs;
-        private List<Action> actions;
+        private ChecksConclusion conclusion;
+        private Output output;
```
XiongKezhi, line 97: I'm too careless, I take this field to be a list before  :sweat_smile:

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -88,11 +94,25 @@ public ChecksConclusion getConclusion() {
...
         private List<Action> actions = Collections.emptyList();
 
-        public ChecksDetailsBuilder(final String name, final ChecksStatus status)
-                throws IllegalArgumentException{
+        /**
+         * Construct a builder with the given name and status.
+         *
+         * <p>
+         *     The name will be the same as the check run's name shown on GitHub UI and GitHub uses this name to
+         *     identify a check run, so make sure this name is unique, e.g. "code-coverage".
```
timja, line 105: I think that human readable names are being used for these checks now? Code coverage, rather than code-coverage

but I assume code is implied here, what other coverage would you be interested in?
```suggestion
         *     identify a check run, so make sure this name is unique, e.g. "Coverage".
```

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -118,12 +165,29 @@ public ChecksDetailsBuilder withConclusion(final ChecksConclusion conclusion) {
...
+         * Set the output of a check.
+         *
+         * @param output
+         *         an output of a check
+         * @return this builder
+         * @throws NullPointerException if the <code>outputs</code> is null
+         */
+        public ChecksDetailsBuilder withOutput(final Output output) {
+            // TODO: Should store the clone of the output after output is constructed.
+            Objects.requireNonNull(output);
```
timja, line 178: often it's nice to statically import requireNonNull so that the code read nicer
```suggestion
           requireNonNull(output);
```

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
XiongKezhi: Just copied the name from GitHub doc: https://developer.github.com/v3/checks/runs/#parameters. Either is ok for me. What matters is we may need to find a way to avoid duplicate names.

timja: not our issue IMO, other than documenting it

XiongKezhi:
Any other problems or we can merge it now. @timja @uhafner

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -1,76 +1,212 @@
...
+    @Nullable
     public String getDetailsURL() {
         return detailsURL;
     }
 
-    public String getConclusion() {
+    /**
+     * Returns the conclusion of a check.
+     *
+     * @return the conclusion of a check or null
```
uhafner, line 67: Does `null` mean no conclusion yet? Then it would be better to add an enum instance like for that.

XiongKezhi: Added a `NONE` instance.

uhafner: Then just the JavaDoc should not have `null` anymore.

src/main/java/io/jenkins/plugins/github/checks/api/ChecksDetails.java
```diff
@@ -1,76 +1,212 @@
...
+         *
+         * <p>
+         *     The conclusion should only be set when the <code>status</code> was set {@link ChecksStatus#COMPLETED}
+         *     when constructing this builder.
+         * </p>
+         *
+         * @param conclusion
+         *         the conclusion
+         * @return this builder
+         * @throws NullPointerException if the <code>conclusion</code> is null
```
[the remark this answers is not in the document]
XiongKezhi: Just saw the `Objects.requreNonNull` did: https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-

Sometimes it's just can't find a standard on these the specific points...

Issue #17, opened 2020-06-05
GitHub instance should support okhttp3, proxy and rate limit


XiongKezhi:
Need to refactor the building of `GitHub` instance. See https://github.com/XiongKezhi/checks-api-plugin/pull/8/files?file-filters%5B%5D=.java&hide-deleted-files=true#r435749980

PR #21, opened 2020-06-09
Refactor ChecksPublisher

XiongKezhi:
- [x] Refactor `ChecksPublisher`
- [ ] Add tests for `ChecksPublisher` and `GitHubChecksPublisher`


src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
```diff
@@ -0,0 +1,67 @@
...
+                    .withDetailsURL(context.getRun().getAbsoluteUrl());
+        } catch (IOException e) {
+            throw new IOException("could not publish checks to GitHub", e);
+        }
+
+        // TODO: Add output and Actions, need a strategy mapping the output and actions to the library's
+
+        switch (details.getStatus()) {
+            case QUEUED:
+                builder.withStatus(Status.QUEUED).withStartedAt(new Date()).create();
```
timja, line 51: missed this earlier could we use the java.time APIs, probably a `LocalDate` here?

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
```diff
@@ -1,47 +1,34 @@
...
-        for (ChecksPublisher publisher : jenkins.getExtensionsFor(ChecksPublisher.class)) {
-            if (publisher.autoStatus().contains(ChecksStatus.QUEUED)) {
-                ChecksDetails checks = new ChecksDetailsBuilder(publisher.getName(), ChecksStatus.QUEUED).build();
-                publisher.publishToQueued(run, checks);
-            }
+        ChecksPublisher publisher = ChecksPublisher.fromRun(run);
+        try {
+            publisher.publish(new ChecksDetailsBuilder(CHECKS_NAME, ChecksStatus.QUEUED).build());
+        } catch (IOException e) {
+            e.printStackTrace();
```
timja, line 31: could we at least throw an UncheckedIOException

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
XiongKezhi: But GitHub API library just support `Date`: https://github.com/hub4j/github-api/blob/4f8a64610bb33c687ab530fb181e56031f29327b/src/main/java/org/kohsuke/github/GHCheckRunBuilder.java#L91

timja: ideally we use the modern API, and then we can convert it down, if time allows we should also update github-api

XiongKezhi: maybe just github-api if possible, since here it is the bottom of our plugin and this `GHCheckRunBuilder` is provided directly by github-api.

But github-api uses `Date` everywhere especially in the bottom requesters, when I was implementing part of the `CheckRun` for the github-api before, I tried to use `LocalDate` for this part, but I stopped when realized that...

I'll discuss with Liam again after our main features implemented.

timja: sure is fine then

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @param details
+     *         the details of a check run
+     * @throws IOException if publish the check run failed
+     */
+    @Override
+    public void publish(final ChecksDetails details) throws IOException {
+        GHCheckRunBuilder builder;
+        try {
+            builder = new GitHubBuilder()
```
timja, line 38: can we use the new api in github-branch-source for this?

https://github.com/XiongKezhi/checks-api-plugin/issues/17

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
```diff
@@ -0,0 +1,67 @@
...
+                break;
+            case COMPLETED:
+                builder.withCompletedAt(new Date())
+                        .withConclusion(Conclusion.SUCCESS) // TODO: need a strategy mapping the conclusion
+                        .create();
+        }
+    }
+
+    @Override
+    protected Optional<ChecksPublisher> createPublisher(final ChecksContext context) {
```
timja, line 64: I don't understand why this is optional?

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
```diff
@@ -1,74 +1,75 @@
...
-    public default void publishToCompleted (final Run<?, ?> run, final ChecksDetails details) {
-        ChecksClient.completeCheckRun(run, details);
+    public static class NullChecksPublisher extends ChecksPublisher {
+        @Override
+        public void publish(final ChecksDetails details) {
+        }
+
+        @Override
+        protected Optional<ChecksPublisher> createPublisher(final ChecksContext context) {
+            return Optional.empty();
```
timja, line 72: shouldn't this just return an instance of `NullChecksPublisher` which does nothing?

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
```diff
@@ -1,74 +1,75 @@
...
+import io.jenkins.plugins.github.checks.ChecksContext;
+import io.jenkins.plugins.util.JenkinsFacade;
 
 /**
  * A publisher API for consumers to publish checks.
  */
 @Restricted(Beta.class)
-public interface ChecksPublisher extends ExtensionPoint {
+public abstract class ChecksPublisher implements ExtensionPoint {
+    protected ChecksContext context;
```
uhafner, line 22: Members must be `private`.

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
```diff
@@ -1,74 +1,75 @@
...
-     * Returns the name of a check.
+     * Publishes checks to platforms.
      *
-     * @return the name of a check
+     * @param details
+     *         the details of a check
+     * @throws IOException if publish check failed
      */
-    public abstract String getName();
+    public abstract void publish(final ChecksDetails details) throws IOException;
```
uhafner, line 34: This is a anti-pattern in Jenkins, so you can keep it this way. However, it might make sense to use an unchecked exception since you catch it in your example and rethrow an unchecked exception.

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
```diff
@@ -1,74 +1,75 @@
...
+     * Actually creates a suitable publisher based on the {@code context}.
      *
-     * @return
-     *         Returns the <code>ChecksStatus</code>(s) the publisher wants checks plugin to automatically set.
+     * @param context
+     *         the context of a run
+     * @return a publisher suitable for the context
      */
-    public abstract Set<ChecksStatus> autoStatus();
+    protected abstract Optional<ChecksPublisher> createPublisher(final ChecksContext context);
```
uhafner, line 43: I don't understand how this can work? You create an instance of `ChecksPublisher`. But in order to create one, you already need the instance. I think we need two classes here. One class is used by the consumer and the other class that handles the actual REST api calls to the corresponding SCM providers. 

So `publish` is not abstract, it should work in the following way:
```
ExtensionPoint actualPublisher = getPublisherFor(context);
actualPublisher.publish(); 
```

uhafner: It would be helpful to see a test for the code, then it would be more clear how the API works...

src/main/java/io/jenkins/plugins/github/checks/JobListener.java
```diff
@@ -1,77 +1,66 @@
...
+     * When a job is initializing, creates a check to keep track of the {@code run}.
      */
     @Override
     public void onInitialize(Run run) {
-        for (ChecksPublisher publisher : jenkins.getExtensionsFor(ChecksPublisher.class)) {
-            if (publisher.autoStatus().contains(ChecksStatus.QUEUED)) {
-                ChecksDetails checks = new ChecksDetailsBuilder(publisher.getName(), ChecksStatus.QUEUED).build();
-                publisher.publishToQueued(run, checks);
-            }
+        ChecksPublisher publisher = ChecksPublisher.fromRun(run);
```
uhafner, line 26: Extract common code to a method

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
```diff
@@ -0,0 +1,67 @@
...
+     * @throws IOException if publish the check run failed
+     */
+    @Override
+    public void publish(final ChecksDetails details) throws IOException {
+        GHCheckRunBuilder builder;
+        try {
+            builder = new GitHubBuilder()
+                    .withAppInstallationToken(context.getToken()).build()
+                    .getRepository(context.getRepository())
+                    .createCheckRun(details.getName(), Objects.requireNonNull(context.getHeadSha()))
```
uhafner, line 41: What happens if there is no SHA?

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
```diff
@@ -0,0 +1,67 @@
...
+                    .withDetailsURL(context.getRun().getParent().getAbsoluteUrl() + context.getRun().getNumber() + "/");
```
uhafner, line 42: The URL should be computed in the context

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
```diff
@@ -0,0 +1,67 @@
...
+        } catch (IOException e) {
+            throw new IOException("could not publish checks to GitHub", e);
+        }
+
+        // TODO: Add output and Actions, need a strategy mapping the output and actions to the library's
+
+        switch (details.getStatus()) {
+            case QUEUED:
```
uhafner, line 50: It would be helpful to prefix the enum, so you get not confused with the other ones

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
```diff
@@ -0,0 +1,67 @@
...
+                builder.withStatus(Status.QUEUED).withStartedAt(new Date()).create();
+                break;
+            case IN_PROGRESS:
+                builder.withStatus(Status.IN_PROGRESS).withStartedAt(new Date()).create();
+                break;
+            case COMPLETED:
+                builder.withCompletedAt(new Date())
+                        .withConclusion(Conclusion.SUCCESS) // TODO: need a strategy mapping the conclusion
+                        .create();
```
uhafner, line 59: What happens with the created object?

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
XiongKezhi: The SHA is used to create a `CheckRunBuilder` in GitHub API, and the `create()` in `CheckRunBuilder` actually creates a check run on GitHub. See https://github.com/hub4j/github-api/blob/992d2b937c3509aeaf1e1ea7476595cf369fe4e7/src/main/java/org/kohsuke/github/GHRepository.java#L1915

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
XiongKezhi: Yes, I should encapsulate that logic..

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
XiongKezhi: Intend to inherit this to sub-classes, but anyway, it seems this class need to be refacorted

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
XiongKezhi: The code in `JobListener` shows how to use the API, I'll add test later since it's kind of integration test, so would be more complicate.

src/main/java/io/jenkins/plugins/github/checks/api/ChecksPublisher.java
XiongKezhi: Will refactor this part

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
XiongKezhi: This method will call the GitHub's REST API,  and will represent the response as a `GitHubCheckRun` object, I guess now we don't need the returned object. See https://github.com/hub4j/github-api/blob/master/src/main/java/org/kohsuke/github/GHCheckRun.java

src/test/java/io/jenkins/plugins/github/checks/GitHubChecksPublisherTest.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * 1. unit test: we just need to mock the parameters of the method and verify the returned {@link GHCheckRunBuilder}.
+ * However, I haven't found a way to successfully verify the fields of it since its an external class from github-api.
+ *
+ * 2. integration test: we need to construct the parameters very carefully, since they have to be the parameters
+ * that will appear in real practises. By this way, we just need to verify the returned values of the method which is a
+ * {@link GHCheckRun} instance.
+ */
+@Disabled
+class GitHubChecksPublisherTest {
```
XiongKezhi, line 28: Need some help here! See JavaDoc.

src/main/java/io/jenkins/plugins/github/checks/ChecksConclusion.java
```diff
@@ -1,12 +1,24 @@
...
+    NONE(null),
+    ACTION_REQUIRED(Conclusion.ACTION_REQUIRED),
+    SKIPPED(Conclusion.CANCELLED), // TODO: need a PR for GitHub API to add that
+    CANCELED(Conclusion.CANCELLED),
+    TIME_OUT(Conclusion.TIMED_OUT),
+    FAILURE(Conclusion.FAILURE),
+    NEUTRAL(Conclusion.NEUTRAL),
+    SUCCESS(Conclusion.SUCCESS);
+
+    private final Conclusion CheckRunConclusion;
```
timja, line 15: shouldn't it be this for java naming standards?
```suggestion
    private final Conclusion checkRunConclusion;
```

src/main/java/io/jenkins/plugins/github/checks/ChecksConclusion.java
```diff
@@ -1,12 +1,24 @@
...
+
+    ChecksConclusion(final Conclusion checkRunConclusion) {
+        this.CheckRunConclusion = checkRunConclusion;
+    }
+
+    public Conclusion toCheckRunConclusion() {
+        return CheckRunConclusion;
```
timja, line 22:
```suggestion
        return checkRunConclusion;
```

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
```diff
@@ -0,0 +1,57 @@
...
+     * Publishes a GitHub check run.
+     *
+     * @param details
+     *         the details of a check run
+     * @throws IOException if publish the check run failed
+     */
+    @Override
+    public void publish(final ChecksDetails details) throws IOException {
+        // TODO: use the github url in the project's configuration
+        GitHub gitHub = Connector.connect(GITHUB_URL, context.getCredential());
```
timja, line 33: the url should be available on the credential

src/test/java/io/jenkins/plugins/github/checks/GitHubChecksPublisherTest.java
```diff
@@ -0,0 +1,69 @@
...
+        GitHub gitHub = createGitHub();
+
+        ChecksDetails details = mock(ChecksDetails.class);
+        when(details.getName()).thenReturn(CHECK_NAME);
+        when(details.getStatus()).thenReturn(ChecksStatus.QUEUED);
+        when(details.getDetailsURL()).thenReturn(null);
+        when(details.getConclusion()).thenReturn(ChecksConclusion.NONE);
+
+        GitHubChecksPublisher publisher = new GitHubChecksPublisher(context);
+        GHCheckRunBuilder builder = publisher.createBuilder(gitHub, details, context);
```
timja, line 65: looks like you can't verify the fields there and also there's no equals method so can't check equality

src/test/java/io/jenkins/plugins/github/checks/GitHubChecksPublisherTest.java
```diff
@@ -0,0 +1,69 @@
...
+import static org.mockito.Mockito.*;
+
+/**
+ * This test class hasn't been determined yet, since we can test {@link GitHubChecksPublisher#createBuilder} in unit
+ * test or integration test, but I meet difficulties in any way:
+ *
+ * 1. unit test: we just need to mock the parameters of the method and verify the returned {@link GHCheckRunBuilder}.
+ * However, I haven't found a way to successfully verify the fields of it since its an external class from github-api.
+ *
+ * 2. integration test: we need to construct the parameters very carefully, since they have to be the parameters
```
timja, line 23: ideally it's something that can run once and be recorded and then run over again without contacting github

XiongKezhi: I'm setting up a mock server using WireMock today, it's a little complex, may take a while.

I'll post the code right after I finished.

src/test/java/io/jenkins/plugins/github/checks/GitHubChecksPublisherTest.java
uhafner: The author of `org.kohsuke.github` did everything to prevent users of the API to test their code 🤷. I think it is not only the missing getters it is also that all classes are final which makes stubbing them very difficult. This library is a good example what happens if one is not using TDD right from the beginning. 

I think just focus on the other tests and come back later. Since the type is deprecated anyway: do we need to replace that part with something else?

timja: every preview api in github-api is marked as deprecated, they do it so that you noticed you're using a preview api, bit weird...

src/main/java/io/jenkins/plugins/github/checks/api/ChecksOutput.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.Objects;
+
+import org.kohsuke.github.GHCheckRun;
+import org.kohsuke.github.GHCheckRunBuilder.Output;
+
+/**
+ * A simple wrapper for {@link GHCheckRun.Output} to create an output for the GitHub's check run.
+ */
+public class ChecksOutput {
```
XiongKezhi, line 11: Changed the class name from `Output` to `ChecksOutput` just to distinguish it from the `Output` class from github-api. Not sure if this is the right way to wrap the `Output` class from github-api, but this is a simple way that needs the least code.

Problems exist here like the fields of the `Output` like `title` and `summary` are invisible (no getters), but if we assume that the getters in `ChecksDetails` will not be used by users, the same assumption works here; and tests for this class becomes harder as well.

src/test/java/io/jenkins/plugins/github/checks/api/ChecksDetailsTest.java
```diff
@@ -32,21 +32,21 @@ void shouldCreateWhenBuildOnlyWithConstructor() {
     @Test
     void shouldCreateWhenBuildWithAllFields() {
         // TODO: Changes may required here after refactoring Output and Action
-        Output output = new Output();
+        // ChecksOutputBuilder outputBuilder = new ChecksOutput();
```
XiongKezhi, line 35: Comment for now, will refactor after determing the structure of `ChecksOutput` class.

src/main/java/io/jenkins/plugins/github/checks/api/ChecksOutput.java
uhafner: I think you model should not depend on `Output`. It should be a simple bean (well actually a builder) that collects the information from the consumer. Then we have a class in the GitHub package that converts the `ChecksOutput` to the `Output`. (The same would happen in a similar GitLab plugin).

src/main/java/io/jenkins/plugins/github/checks/api/ChecksAnnotation.java
```diff
@@ -0,0 +1,70 @@
...
+ * A simple wrapper for {@link GHCheckRunBuilder.Annotation} to create an annotation for the GitHub's check run.
+ */
+public class ChecksAnnotation {
+    private final Annotation annotation;
+
+    private ChecksAnnotation(final Annotation annotation) {
+        this.annotation = annotation;
+    }
+
+    public Annotation toGitHubChecksAnnotation() {
```
uhafner, line 17: see below, should be stand alone.

src/main/java/io/jenkins/plugins/github/checks/api/ChecksOutput.java
XiongKezhi: So that also means we also need converters for `ChecksConclusion` and `ChecksStatus`

uhafner: Yes. I think you already provided such a thing in this PR. The only problem I see is that you mixed the API class with the GitHub class. 

Maybe it makes sense to move the API to a different package and then define an architecture test that GitHub depends on Api (and not the other way round). Then such design problems will show up immediately. 

So I think we would have a `GitHubAdapter` that maps your API to the corresponding GitHub calls (or GitHub objects). This could be done in one or more classes, I'm not sure yet. Start simple in one class, then we can split if required.

Also it would make sense to have a look in the beginning on how the consumer will use the API (see example in https://github.com/XiongKezhi/checks-api-plugin/blob/6b9013e548632478297f5914caacd0cb28d02428/src/main/java/io/jenkins/plugins/WarningsChecks.java#L18)

XiongKezhi: One problem I met when I'm progressing today:

Since currently, we don't have other platform use-cases besides the GitHub, so the fields in `CheckOutput` will be the same as the ones in `Output`. And even if there are other platforms usage later, we have to add their unique fields to `ChecksOutput`. For example, if the GitLab requires a parameter (say `creator`) that is not needed by GitHub, we need to add this fields and the corresponding getters and setters. I wonder if there is way that can release us from this future works.

uhafner: I think we can't. Even GitHub might add new elements in future versions of the API. The only thing we can do here: use abstract base classes for such elements (and not interfaces), then we can provide a smooth upgrade guide for the consumers. 

(And if we find some time at the end of this project, we can still have a quick look at the GitLab API to see how it differs. But for now I think it is save to look at GitHub. I did the same for the Forensics Plugin: it is designed around the semantics of Git. If later on someone provides an implementation for Subversion it might be required to change the API. )

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksDetails.java
```diff
@@ -0,0 +1,91 @@
...
+import org.kohsuke.github.GHCheckRunBuilder.Output;
+import org.kohsuke.github.GHCheckRunBuilder.Annotation;
+
+import io.jenkins.plugins.checks.api.ChecksAnnotation;
+import io.jenkins.plugins.checks.api.ChecksAnnotation.ChecksAnnotationLevel;
+import io.jenkins.plugins.checks.api.ChecksDetails;
+import io.jenkins.plugins.checks.api.ChecksImage;
+import io.jenkins.plugins.checks.api.ChecksOutput;
+
+class GitHubChecksDetails {
```
XiongKezhi, line 16: Not a typical adaptor since I didn't provide a specific interface to adapt to, though maybe this is more straightforward and this class is used internally.

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksDetails.java
```diff
@@ -0,0 +1,91 @@
...
+        // FIXME: should create a lookup table?
+        switch (checksLevel) {
+            case NOTICE:
+                return AnnotationLevel.NOTICE;
+            case FAILURE:
+                return AnnotationLevel.FAILURE;
+            case WARNING:
+                return AnnotationLevel.WARNING;
+            default:
+                return null;
```
timja, line 88: probably want to throw an exception here?

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksDetails.java
```diff
@@ -0,0 +1,91 @@
...
+                .withStartColumn(checksAnnotation.getStartColumn())
+                .withEndColumn(checksAnnotation.getEndColumn());
+    }
+
+    private Image getImage(final ChecksImage checksImage) {
+        return new Image(checksImage.getAlt(), checksImage.getImageUrl()).withCaption(checksImage.getCaption());
+    }
+
+    private AnnotationLevel getAnnotationLevel(final ChecksAnnotationLevel checksLevel) {
+        // FIXME: should create a lookup table?
```
timja, line 79: this should be fine as is I think

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksDetails.java
```diff
@@ -85,7 +84,7 @@ private AnnotationLevel getAnnotationLevel(final ChecksAnnotationLevel checksLev
             case WARNING:
                 return AnnotationLevel.WARNING;
             default:
-                return null;
+                throw new IllegalArgumentException("unsupported checks annotation level");
```
uhafner, line 87: When you throw an exception always add the context. 
```suggestion
                throw new IllegalArgumentException("Unsupported checks annotation level: ", checksLevel);
```

XiongKezhi: Got it

uhafner:
You made a lot of progress since I looked at it, it looks very promising 👍  

I'm not sure if you want us to continuously have a look at the current state of the PR? Or do you ping us with a review request when we should look again at the whole PR?

XiongKezhi:
> 
> 
> You made a lot of progress since I looked at it, it looks very promising 👍
> 
> I'm not sure if you want us to continuously have a look at the current state of the PR? Or do you ping us with a review request when we should look again at the whole PR?

This PR is almost done, just need some minor fixing for coding styles and nits. I'll ping you when I'm totally finished, then we can take a look at the whole and merge it.

src/main/java/io/jenkins/plugins/checks/api/ChecksAnnotation.java
```diff
@@ -0,0 +1,139 @@
...
+        private final int startLine;
+        private final int endLine;
+        private final ChecksAnnotationLevel annotationLevel;
+        private final String message;
+        private int startColumn;
+        private int endColumn;
+        private String title;
+        private String rawDetails;
+
+        public ChecksAnnotationBuilder(final String path,
```
uhafner, line 87: There should be also a constructor with a single line (then start == end). And in the warnings plugin it is also possible to have warnings for the whole file. So providing no start might be a valid option as well?

src/main/java/io/jenkins/plugins/checks/api/ChecksAnnotation.java
```diff
@@ -0,0 +1,139 @@
...
+                final int startLine, final int endLine,
+                final ChecksAnnotationLevel annotationLevel,
```
uhafner, line 89: I would prefer to remove the `level` from the constructor. Then you can use `WARNING` as default (and a `withLevel`).

src/main/java/io/jenkins/plugins/checks/api/ChecksAnnotation.java
XiongKezhi: > And in the warnings plugin it is also possible to have warnings for the whole file.

The `start_line` and `end_line` parameters are required for annotation, so in that way, maybe we should set a start as 0.

src/main/java/io/jenkins/plugins/checks/api/ChecksOutput.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public static class ChecksOutputBuilder {
+        private final String title;
+        private final String summary;
+        private String text;
+        private List<ChecksAnnotation> annotations;
+        private List<ChecksImage> images;
+
+        public ChecksOutputBuilder(final String title, final String summary) {
+            // TODO: determine how GitHub behave when passing a empty String instead of null, possibly use StringUtils
```
XiongKezhi, line 51: Turns out GitHub allows empty strings

src/main/java/io/jenkins/plugins/github/checks/GitHubChecksPublisher.java
XiongKezhi: Can't find a better prefix than 'CHECKS_', but still, this is hard to make itself distinct from the ones from GitHub library since they are so related and almost semantic equivalent. Maybe only the context could help...

XiongKezhi:
I think the initial purpose of this PR is done by now. The tests for `ChecksPublisher` and `GitHubChecksPublisher` need to set up a mock server, so I prefer to keep this PR from those works.

@uhafner @timja kindly review.

src/main/java/io/jenkins/plugins/checks/api/ChecksPublisher.java
```diff
@@ -0,0 +1,31 @@
...
+     */
+    public abstract void publish(final ChecksDetails details) throws IOException;
+
+    public static class NullChecksPublisher extends ChecksPublisher {
+        public NullChecksPublisher(final ChecksContext context) {
+            super(context);
+        }
+
+        @Override
+        public void publish(final ChecksDetails details) {
```
XiongKezhi, line 28: Should we do something here to warn the users or just left it empty?

src/main/java/io/jenkins/plugins/checks/ChecksContext.java
```diff
@@ -0,0 +1,68 @@
...
+
+import edu.hm.hafner.util.VisibleForTesting;
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+
+import org.jenkinsci.plugins.github_branch_source.GitHubAppCredentials;
+import org.jenkinsci.plugins.github_branch_source.GitHubSCMSource;
+import hudson.model.Run;
+import jenkins.scm.api.SCMSource;
+
+// TODO: Refactor this class
```
timja, line 17: what's left here? what's planned for refactoring?

src/main/java/io/jenkins/plugins/checks/ChecksContext.java
```diff
@@ -0,0 +1,68 @@
...
+public class ChecksContext {
+    private final Run<?, ?> run;
+    private final GitHubSCMSource source;
+    private final String headSha;
+
+    private static final Logger LOGGER = Logger.getLogger(ChecksContext.class.toString());
+
+    public ChecksContext(final Run<?, ?> run) {
+        this(run, new ContextResolver());
```
timja, line 26: create a default context resolver which is reused? there's no state here.

src/main/java/io/jenkins/plugins/checks/ChecksContext.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @CheckForNull
+    public GitHubAppCredentials getCredential() {
+        if (source.getCredentialsId() != null) {
+            return CredentialsProvider.findCredentialById(source.getCredentialsId(), GitHubAppCredentials.class, run);
+        }
+        return null;
+    }
+
+    public String getURL() {
```
timja, line 65: documentation? mainly just an example of what the URL looks like so users know if they need to add a slash or not

src/main/java/io/jenkins/plugins/checks/api/ChecksAction.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.Objects;
+
+public class ChecksAction {
+    private final String label;
+    private final String description;
+    private final String identifier;
+
+    public ChecksAction(final String label, final String description, final String identifier) {
+        Objects.requireNonNull(label);
```
timja, line 11: would be nice to static import Objects.requireNonNull here

src/main/java/io/jenkins/plugins/checks/api/ChecksAction.java
```diff
@@ -0,0 +1,35 @@
...
+        Objects.requireNonNull(description);
+        Objects.requireNonNull(identifier);
+
+        this.label = label;
+        this.description = description;
+        this.identifier = identifier;
+    }
+
+    public ChecksAction(final ChecksAction that) {
```
timja, line 20: Why do we need this? it's only used in the tests

src/main/java/io/jenkins/plugins/checks/api/ChecksAnnotation.java
```diff
@@ -0,0 +1,146 @@
...
+        return rawDetails;
+    }
+
+    public enum ChecksAnnotationLevel {
+        NOTICE,
+        WARNING,
+        FAILURE
+    }
+
+    public static class ChecksAnnotationBuilder {
```
timja, line 81: missing docs in this class

src/main/java/io/jenkins/plugins/checks/api/ChecksAnnotation.java
```diff
@@ -0,0 +1,146 @@
...
+        private final ChecksAnnotationLevel annotationLevel;
+        private final String message;
+        private int startColumn;
+        private int endColumn;
+        private String title;
+        private String rawDetails;
+
+        public ChecksAnnotationBuilder(final String path, final int startLine, final int endLine,
+                final ChecksAnnotationLevel annotationLevel, final String message) {
+            Objects.requireNonNull(path);
```
timja, line 94: static import would be nice

src/main/java/io/jenkins/plugins/checks/api/ChecksAnnotation.java
```diff
@@ -0,0 +1,146 @@
...
+                throw new IllegalArgumentException("startLine and endLine attributes should the same "
+                        + "when adding column");
+            }
+            this.startColumn = startColumn;
+            return this;
+        }
+
+        public ChecksAnnotationBuilder withEndColumn(final int endColumn) {
+            if (startLine != endLine) {
+                throw new IllegalArgumentException("startLine and endLine attributes should the same "
```
timja, line 121:
```suggestion
                throw new IllegalArgumentException("startLine and endLine attributes must be the same "
```

src/main/java/io/jenkins/plugins/checks/api/ChecksConclusion.java
```diff
@@ -0,0 +1,5 @@
+package io.jenkins.plugins.checks.api;
+
+public enum ChecksConclusion {
+    NONE, ACTION_REQUIRED, SKIPPED, CANCELED, TIME_OUT, FAILURE, NEUTRAL, SUCCESS
```
timja, line 4: docs would be good, or at least a link to the github statuses

src/main/java/io/jenkins/plugins/checks/api/ChecksDetails.java
```diff
@@ -191,9 +192,9 @@ public ChecksDetailsBuilder withOutput(final Output output) {
          * @return this builder
          * @throws NullPointerException if the {@code actions} is null
          */
-        public ChecksDetailsBuilder withActions(final List<Action> actions) {
+        public ChecksDetailsBuilder withActions(final List<ChecksAction> actions) {
             requireNonNull(actions);
-            this.actions = Collections.unmodifiableList(actions);
+            this.actions = Collections.unmodifiableList(new ArrayList<>(actions));
```
timja, line 197: unmodifiable list returns a copy of the list anyway, no need to double copy
```suggestion
            this.actions = Collections.unmodifiableList(actions);
```

src/main/java/io/jenkins/plugins/checks/api/ChecksImage.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Objects;
+
+public class ChecksImage {
+    private final String alt;
+    private final String imageUrl;
+    private String caption;
+
+    public ChecksImage(final String alt, final String imageUrl) {
+        Objects.requireNonNull(alt);
```
timja, line 11: static import would be nice

src/main/java/io/jenkins/plugins/checks/api/ChecksImage.java
```diff
@@ -0,0 +1,39 @@
+package io.jenkins.plugins.checks.api;
+
+import java.util.Objects;
+
+public class ChecksImage {
```
timja, line 5: docs would be good

src/main/java/io/jenkins/plugins/checks/api/ChecksOutput.java
```diff
@@ -0,0 +1,86 @@
+package io.jenkins.plugins.checks.api;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+public class ChecksOutput {
```
timja, line 8: docs would be good

src/main/java/io/jenkins/plugins/checks/api/ChecksPublisher.java
timja: Probably not needed for now, we can add later if it's an issue

src/main/java/io/jenkins/plugins/checks/api/ChecksPublisherFactory.java
```diff
@@ -0,0 +1,42 @@
...
+import io.jenkins.plugins.checks.ChecksContext;
+import io.jenkins.plugins.checks.api.ChecksPublisher.NullChecksPublisher;
+import io.jenkins.plugins.util.JenkinsFacade;
+
+/**
+ * A publisher API for consumers to publish checks.
+ */
+@Restricted(Beta.class)
+public abstract class ChecksPublisherFactory implements ExtensionPoint {
+    protected abstract Optional<ChecksPublisher> createPublisher(final ChecksContext context);
```
timja, line 20: docs?

src/main/java/io/jenkins/plugins/checks/api/ChecksPublisherFactory.java
```diff
@@ -0,0 +1,42 @@
...
+        return findAllPublisherFactories().stream()
+                .map(factory -> factory.createPublisher(context))
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .findFirst()
+                .orElse(new NullChecksPublisher(context));
+    }
+
+    private static List<ChecksPublisherFactory> findAllPublisherFactories() {
+        return new JenkinsFacade().getExtensionsFor(ChecksPublisherFactory.class);
```
timja, line 40: shouldn't there be a way to inject the JenkinsFacade?  possibly via a setter here?

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksDetails.java
timja: minimal docs would be good

src/test/java/io/jenkins/plugins/checks/api/ChecksAnnotationTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+import io.jenkins.plugins.checks.api.ChecksAnnotation.ChecksAnnotationBuilder;
+import io.jenkins.plugins.checks.api.ChecksAnnotation.ChecksAnnotationLevel;
+
+import static io.jenkins.plugins.checks.api.ChecksAnnotationAssert.*;
+import static org.assertj.core.api.Assertions.*;
+
+class ChecksAnnotationTest {
+    private final static String PATH =
+            "github-checks-api-pluginsrc\\main\\java\\io\\jenkins\\plugins\\checks\\CheckGHEventSubscriber.java";
```
timja, line 13: missing / after plugin

can't you use a forward /? It would mean no escaping needed
```suggestion
            "github-checks-api-plugin\\src\\main\\java\\io\\jenkins\\plugins\\checks\\CheckGHEventSubscriber.java";
```

src/test/java/io/jenkins/plugins/checks/api/ChecksOutputTest.java
```diff
@@ -0,0 +1,57 @@
...
+    private final static String TITLE = "Coverage Report";
+    private final static String SUMMARY = "All code have been covered";
+
+    @Test
+    void shouldBuildCorrectlyWithOnlyRequiredFields() {
+        final ChecksOutput checksOutput = new ChecksOutputBuilder(TITLE, SUMMARY).build();
+
+        assertThat(checksOutput).hasTitle(TITLE)
+                .hasSummary(SUMMARY)
+                .hasChecksAnnotations(Collections.emptyList())
```
uhafner, line 25:
```suggestion
                .hasNoChecksAnnotations()
```

src/test/java/io/jenkins/plugins/checks/api/ChecksOutputTest.java
```diff
@@ -0,0 +1,57 @@
...
+    void shouldBuildCorrectlyWithAllFields() {
+        final String text = "#Markdown Supported Text";
+        final List<ChecksAnnotation> annotations =
+                Arrays.asList(mock(ChecksAnnotation.class), mock(ChecksAnnotation.class), mock(ChecksAnnotation.class));
+        final List<ChecksImage> images =
+                Arrays.asList(mock(ChecksImage.class), mock(ChecksImage.class));
+
+        final ChecksOutput checksOutput = new ChecksOutputBuilder(TITLE, SUMMARY)
+                .withText(text)
+                .withAnnotations(annotations)
```
uhafner, line 39: When we actually use that part in the warnings plugin we will see if this is sufficient or if it would help to have a `addAnnotation` method as well. Maybe it is not required and we can do something like:
```
checksOutput.withAnnotations(
      report.stream()
           .map(warning -> convertWarningToAnnoation(warning))
           .collect(Collectors.toList());
```

src/main/java/io/jenkins/plugins/checks/api/ChecksDetails.java
```diff
@@ -71,21 +70,21 @@ public ChecksConclusion getConclusion() {
     }
 
     /**
-     * Returns the {@link Output} of a check
+     * Returns the {@link ChecksOutput} of a check
      *
-     * @return An {@link Output} of a check or null
+     * @return An {@link ChecksOutput} of a check or null
```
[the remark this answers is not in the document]
XiongKezhi: The `Output` object is not required for a check run and the github-api accept a null value when building a check run.

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksDetails.java
```diff
@@ -0,0 +1,93 @@
...
+            case ACTION_REQUIRED: return Conclusion.ACTION_REQUIRED;
+            default: throw new IllegalArgumentException("unsupported checks conclusion: " + details.getConclusion());
+        }
+    }
+
+    @CheckForNull
+    public Output getOutput() {
+        ChecksOutput checksOutput = details.getOutput();
+        if (checksOutput == null) {
+            return null;
```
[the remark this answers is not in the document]
XiongKezhi: `null` value is acceptable for the github-api when building the check-run, so here it just wants to inherit that.

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksPublisher.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @VisibleForTesting
+    GHCheckRunBuilder createBuilder(final GitHub gitHub, final GitHubChecksDetails details, final ChecksContext context)
+            throws IOException {
+        GHCheckRunBuilder builder = gitHub.getRepository(context.getRepository())
+                .createCheckRun(details.getName(), Objects.requireNonNull(context.getHeadSha()));
+        builder.withStatus(details.getStatus())
+                .withDetailsURL(StringUtils.defaultIfBlank(details.getDetailsURL(), context.getURL()));
+
+        if (details.getOutput() != null) {
```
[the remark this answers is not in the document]
XiongKezhi: The `Output` object contains a list of annotations, comments for code should be there: https://developer.github.com/v3/checks/runs/#output-object

Maybe `<Optional>` is a good way, I'll try to refactor.

src/main/java/io/jenkins/plugins/checks/ChecksContext.java
XiongKezhi: The fields in this class like `GitHubSCMSource` and `GitHubAppCredentials` all have generic parent `SCMSource` and `StandardUsernamePasswordCredentials`, I'll refactor them and also avoid return the `null` values but throws exceptions.

src/main/java/io/jenkins/plugins/checks/api/ChecksAnnotation.java
XiongKezhi: The annotation level is a required parameter for check run: https://developer.github.com/v3/checks/runs/#annotations-object. Since we are building the general checks based on this semantic meaning, I think it is makes more sense to ask the users for the required parameters.

src/main/java/io/jenkins/plugins/checks/ContextResolver.java
```diff
@@ -2,44 +2,55 @@
...
 import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMRevision;
 import jenkins.scm.api.SCMSource;
 
-class ContextResolver {
-    @CheckForNull
-    public GitHubSCMSource resolveSource(final Run<?, ?> run) {
-        return (GitHubSCMSource) SCMSource.SourceByItem.findSource(run.getParent());
+public class ContextResolver {
+    @NonNull
```
timja, line 15: isn't nonnull set by default at the package level?

XiongKezhi: Yes, just I haven't get used to that and will remove it immediately.

src/main/java/io/jenkins/plugins/checks/github/GitHubContextResolver.java
```diff
@@ -0,0 +1,58 @@
...
+     *         a run of a GitHub Branch Source project
+     * @return the resolved source
+     */
+    @NonNull
+    public GitHubSCMSource resolveSource(final Run<?, ?> run) {
+        SCMSource source = super.resolveSource(run);
+        if (source instanceof GitHubSCMSource) {
+            return (GitHubSCMSource) source;
+        } else {
+            throw new IllegalStateException("The scm source of the run is not an instance of GitHubSCMSource: "
```
XiongKezhi, line 28: Threw the `illegalStateException` because if the run doesn't contain a `GitHubSCMSource`, then the `GitHubChecksPublisher` publisher shouldn't be used and the code will never get here.

XiongKezhi:
I removed the generic `ChecksContext` for simplicity but reserved `ContextResolver` as a utility class.

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksPublisher.java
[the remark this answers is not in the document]
XiongKezhi: I reconsidered the `null` problem for the checks parameters part: if we use `Optional`, all the non-required parameters for a GitHub check run are `Optional` and when we pass these parameters to github library, we have to invoke `isPresent` for each one of them before `get` the value, so that will make the code to be much more complex, while the github library accepts and will check null, so I'm not sure whether it is still worth using `Optional` for these parameters.

src/main/java/io/jenkins/plugins/checks/api/ChecksAction.java
XiongKezhi: I kept this method for the deep copy of the actions.
https://github.com/XiongKezhi/checks-api-plugin/blob/c97fb6b2f021786c2bddbf9f1ff10b23e6014c56/src/main/java/io/jenkins/plugins/checks/api/ChecksDetails.java#L196-L200

XiongKezhi:
Merged this PR for splitting the works, the left problems in the unresolved conversations are all unrelated to this PR, I'll keep resolving them but in other PR or commits.

PR #22, opened 2020-06-26
Add mock server test for GitHub check run publishing

XiongKezhi:
Using [Wiremock](http://wiremock.org/) for server mock test. 

Currently using the request and response records from github-api library: https://github.com/hub4j/github-api/tree/master/src/test/resources/org/kohsuke/github/GHCheckRunBuilderTest/wiremock/createCheckRun, will eventually replace them with the ones recorded ourselves.


src/main/java/io/jenkins/plugins/checks/github/GitHubChecksPublisher.java
```diff
@@ -62,12 +62,12 @@ GHCheckRunBuilder createBuilder(final GitHub gitHub, final GitHubChecksDetails d
...
 
-        if (details.getStatus() == Status.COMPLETED) {
-            builder.withConclusion(details.getConclusion());
-            builder.withCompletedAt(new Date());
-        } else {
-            builder.withStartedAt(new Date());
-        }
+        builder.withConclusion(details.getConclusion());
+
+        builder.withStartedAt(Date.from(details.getStartedAt().atZone(ZoneId.systemDefault()).toInstant()));
```
timja, line 67: is the system default zone the best? should it be UTC instead?

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksPublisher.java
```diff
@@ -62,12 +62,12 @@ GHCheckRunBuilder createBuilder(final GitHub gitHub, final GitHubChecksDetails d
             builder.add(details.getOutput());
         }
 
-        if (details.getStatus() == Status.COMPLETED) {
-            builder.withConclusion(details.getConclusion());
-            builder.withCompletedAt(new Date());
-        } else {
-            builder.withStartedAt(new Date());
-        }
+        builder.withConclusion(details.getConclusion());
```
timja, line 65: should these builders be chained?
```
builder.withConclusion()
  .withStartedAt()
  .withCompletedAt()
```

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksPublisher.java
XiongKezhi: Yes, UTC is better.
https://github.com/hub4j/github-api/blob/8fd827040b29471f9fad942344e22cf8f5906e3c/src/main/java/org/kohsuke/github/GitHubClient.java#L702

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksPublisher.java
```diff
@@ -54,20 +54,18 @@ public void publish(final ChecksDetails details) {
...
             final GitHubChecksContext context) throws IOException {
         GHCheckRunBuilder builder = gitHub.getRepository(context.getRepository())
-                .createCheckRun(details.getName(), context.getHeadSha());
-        builder.withStatus(details.getStatus())
-                .withDetailsURL(StringUtils.defaultIfBlank(details.getDetailsURL(), context.getURL()));
+                .createCheckRun(details.getName(), context.getHeadSha())
+                .withStatus(details.getStatus())
+                .withDetailsURL(StringUtils.defaultIfBlank(details.getDetailsURL(), context.getURL()))
+                .withConclusion(details.getConclusion())
+                .withStartedAt(Date.from(details.getStartedAt().atZone(ZoneId.of("UTC")).toInstant()))
```
timja, line 61: there's a constant for this `ZoneOffset.UTC`

XiongKezhi:
The mock server test works fine in my local environment, I'll figure out why it fails on GitHub Actions and fix it.

PR #23, opened 2020-06-27
An experiment to use optional for parameters

XiongKezhi:
- [x] Moved the GitHub's validation checks from the generic `ChecksDetails` to `GitHubChecksDetails`
- [x] Use `Optional` for parameters in `ChecksDetails`
- [x] Refactor validation checks and use `Optional` for other generic api objects like `ChecksAnnotation`
- [x] Refactor tests
- [x] Add JavaDocs

But I'm not sure if this is a better way or necessary for the validation or null checks, need some opinions so I can decide whether to keep progressing.


src/main/java/io/jenkins/plugins/checks/github/GitHubChecksDetails.java
```diff
@@ -139,6 +167,29 @@ public Output getOutput() {
     }
 
     private Action getAction(final ChecksAction checksAction) {
+        if (details.getActions().size() > 3) {
```
XiongKezhi, line 170: Not sure if it's necessary to do such validation checks.

Pros: Handle the wrong cases immediately without posting to GitHub; easier for the consumers to locate why a check run fails to publish.

Cons: Will have many trivial checks; may change when GitHub's rules change

timja: I don’t recommend specific validation.
Just let GitHub validate it and make sure we present the error nicely

XiongKezhi: Got it

src/main/java/io/jenkins/plugins/checks/api/ChecksDetails.java
```diff
@@ -123,30 +117,43 @@ public ChecksOutput getOutput() {
...
-         * @param status
-         *         the status which the check run will be set
-         *
-         * @throws IllegalArgumentException if the name is blank or the status is null
+         *         the check's name
+         * @return this builder
+         * @throws NullPointerException if the {@code name} is null
          */
-        public ChecksDetailsBuilder(final String name, final ChecksStatus status) {
-            if (StringUtils.isBlank(name)) {
```
uhafner, line 141: Why did you remove this? Does it make sense to have a check with blanks only?

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksDetails.java
```diff
@@ -38,6 +40,9 @@
      * @param details the details of a generic check run
      */
     public GitHubChecksDetails(final ChecksDetails details) {
+        if (details.getStatus() == ChecksStatus.COMPLETED && details.getConclusion() == ChecksConclusion.NONE) {
+            throw new IllegalArgumentException("The conclusion is null when status is completed.");
```
uhafner, line 44:
```suggestion
            throw new IllegalArgumentException("No conclusion has been set when status is completed.");
```

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksDetails.java
```diff
@@ -47,22 +52,30 @@ public GitHubChecksDetails(final ChecksDetails details) {
      * @return the name of the check
      */
     public String getName() {
-        return details.getName();
+        return details.getName()
+                .filter(name -> !StringUtils.isBlank(name))
```
uhafner, line 56:
```suggestion
                .filter(StringUtils::isNotBlank)
```

src/main/java/io/jenkins/plugins/checks/github/GitHubChecksDetails.java
```diff
@@ -109,27 +138,26 @@ public Conclusion getConclusion() {
...
-        if (checksOutput == null) {
-            return null;
+    public Optional<Output> getOutput() {
+        Output output = null;
+        if (details.getOutput().isPresent()) {
+            ChecksOutput checksOutput = details.getOutput().get();
+            output = new Output(checksOutput.getTitle(), checksOutput.getSummary())
+                    .withText(checksOutput.getText());
+            checksOutput.getChecksAnnotations().stream().map(this::getAnnotation).forEach(output::add);
+            checksOutput.getChecksImages().stream().map(this::getImage).forEach(output::add);
```
uhafner, line 157:
```suggestion
    public Optional<Output> getOutput() {
        if (details.getOutput().isPresent()) {
            ChecksOutput checksOutput = details.getOutput().get();
            Output output = new Output(checksOutput.getTitle(), checksOutput.getSummary())
                    .withText(checksOutput.getText());
            checksOutput.getChecksAnnotations().stream().map(this::getAnnotation).forEach(output::add);
            checksOutput.getChecksImages().stream().map(this::getImage).forEach(output::add);
            return Optional.of(output);
        }
        return Optional.empty();
```

src/main/java/io/jenkins/plugins/checks/api/ChecksDetails.java
XiongKezhi: I moved all these validations from the `api` to `github` package, I think maybe this is a better design that we delegate these specific works to the specific adapters (say `GitHubChecksDetails`). Hence, once there is a change in the GitHub API's restrictions (except adding or removing parameters), we don't have to change this API part which may be later an individual plugin.

So, this blank check has been moved to:
https://github.com/XiongKezhi/checks-api-plugin/blob/4db4712590eb3a40c10666829f6c975b611c8a3f/src/main/java/io/jenkins/plugins/checks/github/GitHubChecksDetails.java#L56

src/main/java/io/jenkins/plugins/checks/ContextResolver.java
```diff
@@ -22,8 +22,7 @@ public String resolveHeadSha(final SCMSource source, final Run<?, ?> run) {
         try {
             return resolveHeadSha(source.fetch(head, null));
         } catch (IOException | InterruptedException e) {
-            throw new IllegalStateException(String.format("Could not resolve head sha, source: %s, run: %s",
-                    source, run));
+            throw new IllegalStateException("Could not resolve head sha, source: " + e);
```
timja, line 25: Is there a reason you aren't logging the exception? in the above code you're calling toString implicitly on it and getting the message
```suggestion
            throw new IllegalStateException("Could not resolve head sha, source: ", e);
```

src/test/java/io/jenkins/plugins/checks/github/GitHubChecksDetailsTest.java
```diff
@@ -15,15 +17,17 @@
...
                 .withConclusion(ChecksConclusion.SUCCESS)
                 .withDetailsURL("https://ci.jenkins.io")
                 .build();
 
         GitHubChecksDetails gitHubDetails = new GitHubChecksDetails(details);
         assertThat(gitHubDetails.getName()).isEqualTo("checks");
         assertThat(gitHubDetails.getStatus()).isEqualTo(Status.COMPLETED);
-        assertThat(gitHubDetails.getConclusion()).isEqualTo(Conclusion.SUCCESS);
-        assertThat(gitHubDetails.getDetailsURL()).isEqualTo("https://ci.jenkins.io");
+        assertThat(gitHubDetails.getConclusion()).isEqualTo(Optional.of(Conclusion.SUCCESS));
```
timja, line 30: this isn't really how you're supposed to use matchers, try this instead:
https://github.com/spotify/java-hamcrest

src/main/java/io/jenkins/plugins/checks/ContextResolver.java
XiongKezhi: No... I got something useful from the e's message just now and so eagerly want to print it that I forget to keep the whole original exception.

src/main/java/io/jenkins/plugins/checks/api/ChecksAnnotation.java
```diff
@@ -91,76 +96,107 @@ public String getRawDetails() {
...
+
+        /**
+         * Sets the start line of annotation
+         *
          * @param startLine
-         *         the start line of the annotation
+         *         the start line of code to annotate
+         * @return this builder
+         */
+        public ChecksAnnotationBuilder withStartLine(final Integer startLine) {
```
uhafner, line 149: Why do you allow `Integer` when you throw an error on `null`? The same thing is the field that is now an `Integer`. What is the reason for changing this?

src/main/java/io/jenkins/plugins/checks/api/ChecksDetails.java
```diff
@@ -251,41 +248,37 @@ public ChecksDetailsBuilder withOutput(final ChecksOutput output) {
...
+        /**
+         * Adds a {@link ChecksAction}.
+         *
+         * @param action
+         *         the action
+         * @return this builder
+         */
+        public ChecksDetailsBuilder addAction(final ChecksAction action) {
+            requireNonNull(action);
+            if (actions == Collections.EMPTY_LIST) {
```
uhafner, line 267: You should never rely on the implementation of the `emptyList` method. It would be simpler to use new ArrayList in the constructor.

src/main/java/io/jenkins/plugins/checks/api/ChecksOutput.java
```diff
@@ -102,35 +123,61 @@ public ChecksOutputBuilder withText(final String text) {
...
+        /**
+         * Adds a {@link ChecksImage}.
+         *
+         * @param image
+         *         the image
+         * @return this builder
+         */
+        public ChecksOutputBuilder addImage(final ChecksImage image) {
+            requireNonNull(image);
+            if (images == Collections.EMPTY_LIST) {
```
uhafner, line 177: Use `new ArrayList<>();` in constructor.

src/main/java/io/jenkins/plugins/checks/api/ChecksImage.java
```diff
@@ -58,8 +59,8 @@ public String getImageUrl() {
      *
      * @return the short description of the image
      */
-    public String getCaption() {
-        return caption;
+    public Optional<String> getCaption() {
+        return Optional.ofNullable(caption);
     }
```
uhafner, line 64: It is somewhat inconsistent that this simple bean uses a `withCaption` method that indicates a builder. Wouldn't it make more sense to simply add a second constructor with one additional property. Then you can make the instances immutable (and then you do not need to copy them later on in `ChecksOutput`).

src/main/java/io/jenkins/plugins/checks/api/ChecksAnnotation.java
XiongKezhi: Because I removed the constructors with all required parameters to minimize possible changes in the future, I need a state to represent that the user didn't provide this `startLine`. Thus later I can check it: https://github.com/XiongKezhi/checks-api-plugin/blob/fcc5b346d8dc27918000aa28258a68451c628926/src/main/java/io/jenkins/plugins/checks/github/GitHubChecksDetails.java#L200-L201

src/main/java/io/jenkins/plugins/checks/api/ChecksImage.java
XiongKezhi: Making the instances immutable is cool, but for the constructors, I may have to use a constructor with 3 `@Nullable` parameters and return `Optional` in the getters, because if we use a constructor with 2 parameters and another one with 3 parameters, we are like saying "hey, the first two parameters are required by a platform's check (maybe GitHub check run) and the last parameter is optional." That's not what we need in this generic api IMHO.

PR #24, opened 2020-07-04
Ready for hosting and first alpha releasing

XiongKezhi:
We need to host it under jenkinsci and release it before we can consume it in warnings-plugin in phase 2.


pom.xml
```diff
@@ -11,9 +11,10 @@
   </parent>
 
   <groupId>io.jenkins.plugins</groupId>
-  <artifactId>github-checks-api</artifactId>
+  <artifactId>github-checks</artifactId>
   <version>0.1.0-SNAPSHOT</version>
```
XiongKezhi, line 15: What version number should we use now? `0.1.0-alpha` in order to publish to experiment release?

timja: You can choose that at release time just leave it how it is for now

XiongKezhi: ok, I'll try to use the release-drafter by then

